PR #2, opened 2015-11-20
Add a sample Spine application running on Datastore.

AlexanderLitus:


gcd/src/main/java/org/spine3/sample/DatastoreApplication.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.spine3.server.storage.datastore.LocalDatastoreStorageFactory;
+
+/**
+ * A sample application showing basic usage of Spine framework running on GAE local Datastore.
+ *
+ * @author Alexander Litus
+ */
+@SuppressWarnings({"UtilityClassWithoutPrivateConstructor", "UtilityClass"})
+public class DatastoreApplication {
```
alexander-yevsyukov: It is not a Datastore application. It's a sample application, which runs on Datastore storage factory.

gcd/src/main/java/org/spine3/sample/DatastoreApplication.java
```diff
@@ -0,0 +1,44 @@
...
+package org.spine3.sample;
+
+import org.spine3.server.storage.datastore.LocalDatastoreStorageFactory;
+
+/**
+ * A sample application showing basic usage of Spine framework running on GAE local Datastore.
+ *
+ * @author Alexander Litus
+ */
+@SuppressWarnings({"UtilityClassWithoutPrivateConstructor", "UtilityClass"})
```
alexander-yevsyukov: We  have this warnings as smell. What if we extend the Application class replacing the factory method for creating StorageFactory instance?

gcd/src/main/java/org/spine3/sample/DatastoreApplication.java
```diff
@@ -0,0 +1,44 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.spine3.sample;
```
alexander-yevsyukov: How about having a sub-package under `sample` (e.g. `gae`) so that we don't have a name clash?
We also may add some platform-specific classes into this package later. Also, when extracted with all other samples, the user would see nice folder hierarchy, which adds more details as he dives into the framework.

gcd/src/test/java/org/spine3/sample/DatastoreApplicationShould.java
```diff
@@ -0,0 +1,34 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.spine3.sample;
+
+import org.junit.Test;
+
+@SuppressWarnings("InstanceMethodNamingConvention")
```
alexander-yevsyukov: Does it really apply to this test class (in the state it's now)?

gcd/build.gradle
```diff
@@ -5,8 +5,9 @@ repositories {
 
 dependencies {
 
-    // TODO:2015-11-19:alexander.litus: change local dependency to Maven central repository dependency
+    // TODO:2015-11-19:alexander.litus: change local dependencies to Maven central repository dependency
     compile group: 'org.spine3', name: 'core-java', version: '0.1', changing: true, classifier: 'test'
```
alexander-yevsyukov: Why do we have classifier `test` here?

gcd/build.gradle
```diff
@@ -5,8 +5,9 @@ repositories {
...
+    compile group: 'org.spine3', name: 'core-java-sample', version: '0.1', changing: true
```
alexander-yevsyukov: Do we depend on the same version of `core-java` and `core-java-sample`? If yes, consider having a constant.

gcd/build.gradle
```diff
@@ -5,8 +5,9 @@ repositories {
 
 dependencies {
 
-    // TODO:2015-11-19:alexander.litus: change local dependency to Maven central repository dependency
+    // TODO:2015-11-19:alexander.litus: change local dependencies to Maven central repository dependency
```
alexander-yevsyukov: When do we resolve this to-do comment?

gcd/src/main/java/org/spine3/sample/DatastoreApplication.java
AlexanderLitus: OK.

gcd/build.gradle
AlexanderLitus: OK.

gcd/build.gradle
AlexanderLitus: Added a comment:
"...when Spine can be used for evaluation in real projects."

gcd/build.gradle
AlexanderLitus: Because this is the dependency on the test classes (we need to extend them and use in tests).

gcd/src/main/java/org/spine3/sample/DatastoreApplication.java
AlexanderLitus: OK.

gcd/src/main/java/org/spine3/sample/DatastoreApplication.java
AlexanderLitus: Extended and just overrode the constructor to accept only DatastoreStorageFactory.

gcd/src/test/java/org/spine3/sample/DatastoreApplicationShould.java
AlexanderLitus: OK, now it is.

AlexanderLitus:
PTAL.

PR #4, opened 2016-05-27
Add proto definition for Bucket and File

mikhailmikhaylov:
The build does not work, because of outdated version of the project core.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,4 +1,7 @@
 rootProject.name = 'gae-java'
 
+include ':values'
+project(":values").projectDir = "$rootDir/values" as File
+
 include ':gcd'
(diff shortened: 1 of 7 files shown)


alexander-yevsyukov:
@armiol, I guess we may benefit from having such APIs when working with Cloud Storage. But it's not of high priority.

alexander-yevsyukov:
@armiol, @dmdashenkov, please look into this. The code is old, but the idea may be not. Let's discuss.

armiol:
Giving that Google is moving its services to gRPC-based API, the effort of implementing own data description layer (such as the one in this PR) and integrating it with the current Google Cloud Storage client library seems a waste.

Closing as outdated.

PR #9, opened 2016-11-25
Support request fixes

armiol:
Summary of changes:

* Fix ID conversion from `byte[]` to `Message`: use the same mechanism as for `Message` -> `byte[]`. 
* Set BLOB entity fields as unindexed by the Datastore explicitly to avoid 1500 bytes per indexed BLOB restriction.


armiol:
@dmytro-kashcheiev PTAL.

dmytro-kashcheiev (approved):

PR #10, opened 2016-12-07
Format the Datastore `Entity` kind consistently for all entities.

armiol:
Update the naming policy for Datastore `Entity` kinds: use the Java class name instead of `TypeUrl#simpleName` for

 * `EntityStorageRecord`, 
 * `CommandStorageRecord`,
 * `Any` (properties storage).

The same approach is already applied to the rest of Datastore `Entity` kinds.


armiol:
@dmytro-kashcheiev PTAL.

dmytro-kashcheiev (approved):
@armiol LGTM

PR #18, opened 2017-01-13
Improve testing on CI

dmdashenkov:
- Now a Travis build lasts about `3:50` instead of `20 minutes`
 - Now all the consistency awaits happen after the write operations, not before the read operations
 - If `gradlew build` failes, `gradlew codeCoverageReport` does not execute tests once again


dmdashenkov:
@armiol PTAL

armiol (approved):
@dmdashenkov LGTM.

PR #20, opened 2017-01-20
Make Datastore-based `Storage`s extensible

armiol:
Before this PR there was no convenient way to override the default behavior of the `..Storage` implementations.

Summary of changes:
* constructor and key methods/properties are exposed either as `public` or `protected` instead of `private` as before;
* the "extensibility" is covered with tests;
* `gae-java` was updated to the Spine `0.7.9-SNAPSHOT`;
* the own version of the `gae-java` has been increased to `0.7.9-SNAPSHOT` as well.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 project.ext {
     SPINE_VERSION = '0.7.5-SNAPSHOT'
-    GAE_JAVA_VERSION = "0.7.2-SNAPSHOT"
+    GAE_JAVA_VERSION = "0.7.3-SNAPSHOT"
     PROTOBUF_VERSION = '3.1.0'
     SLf4J_VERSION = "1.7.21"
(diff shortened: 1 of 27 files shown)


armiol:
@dmdashenkov PTAL.

build.gradle
```diff
@@ -1,6 +1,6 @@
 project.ext {
-    SPINE_VERSION = '0.7.5-SNAPSHOT'
-    GAE_JAVA_VERSION = "0.7.2-SNAPSHOT"
+    SPINE_VERSION = '0.7.9-SNAPSHOT'
+    GAE_JAVA_VERSION = "0.7.9-SNAPSHOT"
```
dmdashenkov, line 3: Shouldn't we bump the version to the latest (`0.7.24-SNAPSHOT`) at once?

dmdashenkov (approved):
@armiol LGTM
Please, address my single comment

build.gradle
armiol: @dmdashenkov we will, but we'll do that as a next step.

PR #22, opened 2017-02-06
Migrate to core `v0.8.3-SNAPSHOT`

dmdashenkov:
- Support for __archived and deleted__ feature.
 - Support for bulk write in `RecordStorage`.


dmdashenkov:
@alexander-yevsyukov PTAL

gcd/src/main/java/org/spine3/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -55,6 +68,7 @@
     private static final String SNAPSHOT = "SNAPSHOT";
 
     private static final String KIND = AggregateStorageRecord.class.getName();
+    private static final String AGGREGATE_ENTITY_STATUS_KIND = EntityStatus.class.getName();
```
alexander-yevsyukov, line 71: Why do we want FQN instead of short class name?

gcd/src/main/java/org/spine3/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -55,6 +68,7 @@
     private static final String SNAPSHOT = "SNAPSHOT";
```
alexander-yevsyukov, line 68: What does this constant do?

gcd/src/main/java/org/spine3/server/storage/datastore/DatastoreProperties.java
```diff
@@ -51,11 +54,27 @@
     private static final String EVENT_TYPE_PROPERTY_NAME = "event_type";
     private static final String EVENT_ID_PROPERTY_NAME = "event_id";
 
+    private static final String ARCHIVED_PROPERTY_NAME = "archived";
```
alexander-yevsyukov, line 57: Assuming the growing number of constants, I think we need to think of using `enum`s for them. Let's leave it as is for now because we need this update urgently. But let's sit and talk on making this part of the code better.

alexander-yevsyukov (approved):
LGTM, assuming we're going to re-visit this functionality soon. Please see my comments too.

gcd/src/main/java/org/spine3/server/storage/datastore/DsAggregateStorage.java
dmdashenkov: That's how we mark the `AggregateStorageRecord` to be not an event but an aggregate __snapshot__. This string serves as a prefix for all the snapshot IDs in the storage.

gcd/src/main/java/org/spine3/server/storage/datastore/DsAggregateStorage.java
dmdashenkov: This is done to reduce the odds that the framework user will accidentally overlap e.g. `EntityStatus` type, which seems possible for me.

dmytro-kashcheiev (approved):

PR #23, opened 2017-02-07
Clean Up

dmdashenkov:
In this PR:
 - Reduced code issues (like multiple `String` constants etc.).
 - Fixed gradle dependency overlapping.
 - Updated __Spine core__ to `v0.8.7-SNAPSHOT`.
 - Fixed `archived/deleted` implementation.
 - Increased coverage.


dmdashenkov:
@alexander-yevsyukov ,
@dmytro-kashcheiev 
PTAL

gcd/src/main/java/org/spine3/server/storage/datastore/EntityFields.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+package org.spine3.server.storage.datastore;
+
+import com.google.cloud.datastore.Entity;
+
+/**
+ * The container for all the field names we declare in the storage's {@link Entity entities}.
+ *
+ * <p>Represented by a set of enum containers having their {@link Enum#toString toString()} method overloaded.
```
alexander-yevsyukov: Instead of overwriting `toString()` I suggest the following:
1. Have special convention for such enums — they are named the same way the are named in the storage. It would avoid the need for overloading `toString()`.
2. Have an interface that documents this behavior. And make each field enum implement this interface.

gcd/src/main/java/org/spine3/server/storage/datastore/EntityFields.java
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * <p>The values are grouped by their primary usage scope. Though, some of them are used to widely. Those that do are
+ * grouped in a separate enum — {@link CommonFields}.
+ *
+ * @author Dmytro Dashenkov.
+ */
+class EntityFields {
```
alexander-yevsyukov: Having a common interface would allow to avoid the outer class for the enums.

gcd/src/main/java/org/spine3/server/storage/datastore/EntityFields.java
```diff
@@ -0,0 +1,94 @@
...
+
+    enum AggregateFields {
```
alexander-yevsyukov: Why not singular? It would read better: `AggregateField.aggregate_id`.

gcd/src/main/java/org/spine3/server/storage/datastore/EntityFields.java
```diff
@@ -0,0 +1,94 @@
...
+        ARCHIVED,
+        DELETED;
+
+        @Override
+        public String toString() {
+            return name().toLowerCase();
+        }
+    }
+
+    enum CommonFields {
```
alexander-yevsyukov: Maybe just `EntityField`?

gcd/src/main/java/org/spine3/server/storage/datastore/EntityFields.java
```diff
@@ -0,0 +1,94 @@
...
+        CONTEXT_OF_COMMAND,
+        CONTEXT_VERSION;
+
+        @Override
+        public String toString() {
+            return name().toLowerCase();
+        }
+    }
+
+    enum EntityStateFields {
```
alexander-yevsyukov: It should be `EntityStatusField`, I suppose.

gcd/src/main/java/org/spine3/server/storage/datastore/EntityFields.java
```diff
@@ -0,0 +1,94 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package org.spine3.server.storage.datastore;
```
alexander-yevsyukov: Would we benefit from having these field enums somewhere in the `core-java` server part? We may want to have the same thing for the JDBC storage implementation.

gcd/src/test/java/org/spine3/server/storage/datastore/DsAggregateStorageStatusHandlingShould.java
```diff
@@ -0,0 +1,80 @@
...
+import org.spine3.server.aggregate.AggregateStorage;
+import org.spine3.server.aggregate.AggregateStorageStatusHandlingShould;
+import org.spine3.test.aggregate.ProjectId;
+
+/**
+ * @author Dmytro Dashenkov.
+ */
+public class DsAggregateStorageStatusHandlingShould extends AggregateStorageStatusHandlingShould {
+
+    private static final TestDatastoreStorageFactory DATASTORE_FACTORY;
```
alexander-yevsyukov: I think it should be named `dataStoreFactory` instead. Please consider [constant naming guidelines from Google](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names).

alexander-yevsyukov (changes requested):
Please see my comments.

gcd/src/main/java/org/spine3/server/storage/datastore/EntityFields.java
dmdashenkov: I'm afraid this won't help us. `jdbc-storage` has a bit different data model. Also, the string constants representing the entity fields there are contained in package-local classes describing the content of the data table, so they don't make such a mess.

dmdashenkov:
@alexander-yevsyukov PTAL again

gcd/src/main/java/org/spine3/server/storage/datastore/DatastoreProperties.java
```diff
@@ -79,7 +80,7 @@ private DatastoreProperties() {
      */
     static void addTimestampProperty(TimestampOrBuilder timestamp, Entity.Builder entity) {
         final Date date = convertToDate(timestamp);
-        entity.set(TIMESTAMP.toString(), DateTime.copyFrom(date));
+        entity.set(EntityField.timestamp.toString(), DateTime.copyFrom(date));
```
alexander-yevsyukov: Can we have an interface that all the fields would implement? If we pass it as a first parameter to `set()`, we won't need to do `toString()` all the time.

I wonder why doesn't it pick `toString()` automatically...

gcd/src/main/java/org/spine3/server/storage/datastore/field/DatastoreFields.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * <p>Owing to that, we may simply call {@link Enum#toString()} or {@link Enum#name()} to get a valid field name.
+ *
+ * @author Dmytro Dashenkov
+ */
+@Internal
+@Documented
+@Retention(RetentionPolicy.SOURCE)
+@Target(ElementType.TYPE)
+@interface DatastoreFields {
```
alexander-yevsyukov: Does it apply to `Datastore` only?

What I actually meant by “marker interface” was:

```java
public interface StorageField {
}

public enum EntityField implement StorageField {
...
}
````

This way we'd be able to:
1. have custom field implementations that are not based on enums, if we really need to.
2. have common behaviour, if we need

File: gcd/src/main/java/org/spine3/server/storage/datastore/field/AggregateField.java [hunk 10]
```diff
@@ -0,0 +1,38 @@
...
+import org.spine3.server.storage.datastore.DsAggregateStorage;
+
+/**
+ * A container for the {@link Entity} fields specific for the {@link DsAggregateStorage aggregate storage}.
+ *
+ * @author Dmytro Dashenkov
+ * @see DatastoreFields
+ */
+@DatastoreFields
+@Internal
```
alexander-yevsyukov: These are probably `@SPI` because they will be needed for custom storage writers.

alexander-yevsyukov (changes requested):
Please see my comments.

dmdashenkov:
@alexander-yevsyukov PTAL

dmdashenkov:
@armiol PTAL

gcd/src/main/java/org/spine3/server/storage/datastore/DatastoreWrapper.java
```diff
@@ -188,7 +189,7 @@ public Entity read(Key key) {
      * @return results fo the query packed in a {@link List}
      * @see DatastoreReader#run(Query)
      */
-    @SuppressWarnings({"unchecked", "IfStatementWithTooManyBranches", "ChainOfInstanceofChecks"})
+    @SuppressWarnings("unchecked")
```
armiol: Please have a comment why this suppression takes place.

gcd/src/main/java/org/spine3/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -65,6 +65,13 @@
 public class DsAggregateStorage<I> extends AggregateStorage<I> {
 
     private static final String EVENTS_AFTER_LAST_SNAPSHOT_PREFIX = "EVENTS_AFTER_SNAPSHOT_";
+    /**
```
armiol: Please allow an empty line above.

gcd/src/main/java/org/spine3/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -65,6 +65,13 @@
...
+     * Prefix for the string IDs of the {@link AggregateStorageRecord records} which represent an aggregate snapshot,
+     * not an event.
+     *
+     * <p>When reading all the {@link AggregateStorageRecord records} {@link #historyBackward(Object) by an ID},
```
armiol: The last curly brace should be removed.

gcd/src/main/java/org/spine3/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -65,6 +65,13 @@
...
+     * the snapshots will never be included.
```
armiol: `...are never included.`

gcd/src/test/java/org/spine3/server/storage/datastore/DsAggregateStorageStatusHandlingShould.java
```diff
@@ -0,0 +1,80 @@
...
+        datastoreFactory.clear();
+    }
+
+    @AfterClass
+    public static void tearDownClass() {
+        datastoreFactory.tearDown();
+    }
+
+    @Override
+    protected AggregateStorage<ProjectId> getAggregateStorage(Class<? extends Aggregate<ProjectId, ?, ?>> aggregateClass) {
```
armiol: Please split this line into two to fit 100 symbols.

armiol (changes requested):
@dmdashenkov LGTM in general. However there are several issues to be addressed.

dmdashenkov:
@armiol PTAL again

gcd/src/main/java/org/spine3/server/storage/datastore/DatastoreWrapper.java
```diff
@@ -189,7 +189,7 @@ public Entity read(Key key) {
      * @return results fo the query packed in a {@link List}
      * @see DatastoreReader#run(Query)
      */
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings("unchecked") // Untyped query which always works with datastore {@code Entity}.
```
armiol: Please rephrase to make it more user-friendly. Remember, that's an explanation, not a comment to the line where you have that `query` declared.
Also, `datastore` should be a product name.

armiol (changes requested):
@dmdashenkov please see the comment.

armiol (approved):
@dmdashenkov LGTM.

alexander-yevsyukov (approved):
LGTM

PR #25, opened 2017-03-14
Update to Spine 0.8.22

dmdashenkov:
Update to the latest Spine core version should make it easier to migrate to the `StorageFields` support.

In this PR:
 - Migrate to latest Spine core
 - Fix `protobuf-java-lite` dependency issue (for tests)
 - Simplify `DsStandStorage` to avoid ID conversion


dmdashenkov:
@alexander-yevsyukov, PTAL.

alexander-yevsyukov:
> Simplify JDBCStandStorage to avoid ID conversion

@dmdashenkov, did you mean `DsStandStorage`?

Please also note that for abbreviations we capitalize only the first letter. So it should have been `JdbcStandStorage` anyway.

gcd/src/main/java/org/spine3/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -65,26 +67,37 @@
...
      * not an event.
      *
      * The aggregate snapshots are stored under an ID composed from {@code SNAPSHOT} and the aggregate ID.
      */
     private static final String SNAPSHOT = "SNAPSHOT";
 
-    private static final String KIND = AggregateStorageRecord.class.getName();
-    private static final String AGGREGATE_ENTITY_STATUS_KIND = EntityStatus.class.getName();
-    private static final TypeUrl TYPE_URL = TypeUrl.from(AggregateStorageRecord.getDescriptor());
+    private static final String KIND = TypeName.from(AggregateEventRecord.getDescriptor())
```
alexander-yevsyukov: > private static final String KIND

Maybe `AGGREGATE_EVENT_RECORD_KIND`, to make it clearer in usages?

Consider having a constants with the type `TypeName` and use them with `value()` at the places of usage.

gcd/src/main/java/org/spine3/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -65,26 +67,37 @@
...
+                                               .value();
+    private static final String AGGREGATE_LIFECYCLE_KIND = TypeName.from(LifecycleFlags.getDescriptor())
+                                                                   .value();
+    private static final TypeUrl TYPE_URL = TypeUrl.from(AggregateEventRecord.getDescriptor());
```
alexander-yevsyukov: You can have only one instance of `TypeUrl` (instead of `KIND` above) and use conversion methods or `url.getTypeName()` at the place of usage.

gcd/src/main/java/org/spine3/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -285,6 +232,50 @@ protected DsPropertyStorage getPropertyStorage() {
...
+        return Optional.of(flags);
+    }
+
+    @Override
+    public void writeLifecycleFlags(I id, LifecycleFlags flags) {
+        checkNotNull(id);
+        checkNotNull(flags);
+
+        final DatastoreRecordId recordId = generateDatastoreId(id);
+        final Key key = keyFor(datastore, AGGREGATE_LIFECYCLE_KIND, recordId);
```
alexander-yevsyukov: Have a factory method for key construction. We have at least two pieces like this.

gcd/src/main/java/org/spine3/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -199,18 +145,19 @@ protected void writeRecord(I id, AggregateStorageRecord record) {
...
+    protected Iterator<AggregateEventRecord> historyBackward(I id) {
         checkNotNull(id);
 
         final String idString = idToString(id);
         final Query<Entity> query = Query.newEntityQueryBuilder()
-                                    .setKind(KIND)
-                                    .setFilter(StructuredQuery.PropertyFilter.eq(aggregate_id.toString(), idString))
-                                    .build();
+                                         .setKind(KIND)
+                                         .setFilter(StructuredQuery.PropertyFilter.eq(aggregate_id.toString(), idString))
```
alexander-yevsyukov: Can we have a wrapping method for building the filter? We have a number of such code blocks here and there.

gcd/src/main/java/org/spine3/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -232,22 +179,22 @@ public Key apply(@Nullable Entity input) {
 
         final Collection<Entity> filteredEntities = Collections2.filter(eventEntities,
                                                                         new IsActiveAggregateId(inactiveAggregateKeys));
-        final List<AggregateStorageRecord> immutableResult = Entities.entitiesToMessages(filteredEntities, TYPE_URL);
-        final List<AggregateStorageRecord> records = Lists.newArrayList(immutableResult);
+        final List<AggregateEventRecord> immutableResult = Entities.entitiesToMessages(filteredEntities, TYPE_URL);
```
alexander-yevsyukov: > Entities.entitiesToMessages()

Consider renaming the method to `toRecords()`.

gcd/src/main/java/org/spine3/server/storage/datastore/DsEventStorage.java
```diff
@@ -261,11 +283,11 @@ public boolean apply(@Nullable EventStorageRecord event) {
...
     /**
-     * Predicate matching {@linkplain EventFilter EventFilters} to the given {@linkplain EventStorageRecord}.
+     * Predicate matching {@linkplain EventFilter EventFilters} to the given {@linkplain Event}.
      */
     private static class EventFilterChecker implements Predicate<EventFilter> {
 
-        private final EventStorageRecord event;
+        private final Event event;
 
         private static final Function<Any, Message> ANY_UNPACKER = new Function<Any, Message>() {
```
alexander-yevsyukov: The function is already available via `AnyPacker.unpackFunc()`.

gcd/src/main/java/org/spine3/server/storage/datastore/DsRecordStorage.java
```diff
@@ -54,10 +57,6 @@
 import static org.spine3.server.storage.datastore.DatastoreIdentifiers.keyFor;
```
alexander-yevsyukov: Consider having `Ds` prefix for classes like `DsIdentifiers`, `DsProperties`, etc. It's shorter and they are not supposed to be a part of public API anyway.

gcd/src/main/java/org/spine3/server/storage/datastore/DsRecordStorage.java
```diff
@@ -72,26 +71,27 @@
...
     private final DatastoreWrapper datastore;
     private final TypeUrl typeUrl;
+    private final Class<I> idClass;
 
     private static final String VERSION_KEY = "version";
     private static final String TYPE_URL_PROPERTY_NAME = "type_url";
-    private static final TypeUrl RECORD_TYPE_URL = TypeUrl.of(EntityStorageRecord.class);
-    private static final String KIND = EntityStorageRecord.class.getName();
+    private static final TypeUrl RECORD_TYPE_URL = TypeUrl.of(EntityRecord.class);
+    private static final String KIND = TypeName.from(EntityRecord.getDescriptor()).value();
```
alexander-yevsyukov: Consider obtaining a type name from URL.

gcd/src/main/java/org/spine3/server/storage/datastore/Indexes.java
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * A utility for generating the {@linkplain org.spine3.server.storage.Storage#index() storage ID indexes}.
+ *
+ * @author Dmytro Dashenkov.
+ * @see org.spine3.server.storage.Storage#index()
+ */
+public class Indexes {
+
+    private Indexes() {
+    }
```
alexander-yevsyukov: Please insert the following comment block:

```java
// Prevent instantiation of this utility class.
```

This way it would be clear the purpose of having the empty private constructor.

gcd/src/main/java/org/spine3/server/storage/datastore/Indexes.java
```diff
@@ -0,0 +1,108 @@
...
+     * @param filter    custom filters for selecting only a subset of the records of given type from datastore;
+     *                  {@code null} stands for no custom filters
+     * @param <I>       type of the IDs to retrieve
+     * @return an {@link Iterator} of the IDs matching given record kind and filter
+     */
+    public static <I> Iterator<I> indexIterator(DatastoreWrapper datastore,
+                                                String kind,
+                                                Class<I> idClass,
+                                                @Nullable StructuredQuery.Filter filter) {
+        checkNotNull(datastore, "datastore");
```
alexander-yevsyukov: After some further thought, I think we can do without the param names. The information on which param is `null` would be available by stack trace of a through exception. Since we have each check on its own line, we'd be able to find the cause.

alexander-yevsyukov (changes requested):
Please see my comments.

gcd/src/main/java/org/spine3/server/storage/datastore/DsAggregateStorage.java
dmdashenkov: It is a generic method: `static <M extends Message> List<M> entitiesToMessages`
It may be used not only with the `AggregateEventRecords`.
Considering this, I think `entitiesToMessages` is a proper name for the method.

dmdashenkov:
@alexander-yevsyukov PTAL.

alexander-yevsyukov (approved):
LGTM

PR #26, opened 2017-03-29
Datastore `Kind` improvements

dmdashenkov:
Main changes:
 - Each `RecordStorage` has its own `kind` (determined by the record type name)
 - `DsPropertyStorage` uses `Descriptor` to store the values of single type under a single `kind`
 - `DsStandStorage` now uses an instance of `DsStandStorageDelegate extends DsRecordStorage<AggregateStateId>` to delegate its operations.

Now instead of storing the type URL of the record as a field, we store each the `EntityRecord` type under its own type.
Example:

```
class ProjectProjection extends Projection<ProjectId, Project> { /* ... */ }

class TaskProjection extends Projection<TaskId, Task> { /* ... */ }

class ForeignProjectProjection extends Projection<String, Project> { /* ... */ }
```

In this case, records generated from `ProjectProjection` and `ForeignProjectProjection` will be stored under a single Datastore `kind` and the ones from `TaskProjection` will be stored in a separate `kind`.

The `kind` name corresponds the `Entity`'s state type name (defined in protobuf).

### DIFF
--- a/gcd/src/main/java/org/spine3/server/storage/datastore/DsProperties.java
+++ b/gcd/src/main/java/org/spine3/server/storage/datastore/DsProperties.java
@@ -162,7 +162,7 @@ class DsProperties {
         builder.set(producer_id.toString(), producerId);
     }
 
-    static Predicate<Entity> activedEntityPredicate() {
+    static Predicate<Entity> activeEntityPredicate() {
         return NOT_ARCHIVED_OR_DELETED;
     }
 }
--- a/gcd/src/main/java/org/spine3/server/storage/datastore/DsProperties.java
+++ b/gcd/src/main/java/org/spine3/server/storage/datastore/DsProperties.java
@@ -162,7 +162,7 @@ class DsProperties {
         builder.set(producer_id.toString(), producerId);
     }
 
-    static Predicate<Entity> activeEntityPredicate() {
+    static Predicate<Entity> activedEntityPredicate() {
         return NOT_ARCHIVED_OR_DELETED;
     }
 }
--- a/gcd/src/main/java/org/spine3/server/storage/datastore/DsProperties.java
+++ b/gcd/src/main/java/org/spine3/server/storage/datastore/DsProperties.java
@@ -22,7 +22,6 @@ package org.spine3.server.storage.datastore;
 
 import com.google.cloud.datastore.DateTime;
 import com.google.cloud.datastore.Entity;
-import com.google.common.base.Predicate;
 import com.google.protobuf.Message;
 import com.google.protobuf.TimestampOrBuilder;
 import org.spine3.base.Event;
@@ -32,7 +31,6 @@ import org.spine3.protobuf.Messages;
 import org.spine3.server.storage.EntityField;
 import org.spine3.server.storage.LifecycleFlagField;
 
-import javax.annotation.Nullable;
 import java.util.Date;
 
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -57,17 +55,7 @@ import static org.spine3.server.storage.LifecycleFlagField.archived;
 @SuppressWarnings("UtilityClass")
 class DsProperties {
 
-    private static final Predicate<Entity> NOT_ARCHIVED_OR_DELETED = new Predicate<Entity>() {
-        @Override
-        public boolean apply(@Nullable Entity input) {
-            if (input == null) {
-                return false;
-            }
-            final boolean isNotArchived = !isArchived(input);
-            final boolean isNotDeleted = !isDeleted(input);
-            return isNotArchived && isNotDeleted;
-        }
-    };
+
 
     private DsProperties() {
     }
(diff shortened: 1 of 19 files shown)


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@armiol, let's discuss the version number for this incrementation.

gcd/src/test/java/org/spine3/server/storage/datastore/DatastoreStorageFactoryShould.java
```diff
@@ -86,4 +96,10 @@ private TestEntity(String id) {
             super(id);
         }
     }
+
+    private static class  DifferentTestEntity extends AbstractEntity<String, Project> {
```
armiol: Please adjust the formatting.

gcd/src/main/java/org/spine3/server/storage/datastore/DsRecordStorage.java
```diff
@@ -385,8 +386,12 @@ protected void writeRecords(Map<I, EntityRecord> records) {
     @Override
     public Iterator<I> index() {
         checkNotClosed();
-        final StructuredQuery.Filter filter = PropertyFilter.eq(TYPE_URL_PROPERTY_NAME, typeUrl.getTypeName());
-        return Indexes.indexIterator(datastore, KIND, idClass, filter);
+        return Indexes.indexIterator(datastore, kind, idClass);
+    }
+
+    @VisibleForTesting
+    String getKind() {
```
armiol: Let's make a simple value object for `Kind`. It should be immutable and `final`, with a factory method, which validates for `null` etc.

armiol (changes requested):
@dmdashenkov Take a look at the comments. Also, let's make the `getKind()` a `protected` method with a default implementation.

armiol:
@dmdashenkov let's just increase the second minor.

dmdashenkov:
@armiol, PTAL again.

gcd/src/main/java/org/spine3/server/storage/datastore/DsStandStorageDelegate.java
```diff
@@ -0,0 +1,134 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+import static com.google.cloud.datastore.StructuredQuery.PropertyFilter.eq;
+import static org.spine3.server.storage.datastore.DsProperties.activedEntityPredicate;
+import static org.spine3.validate.Validate.isDefault;
+
+/**
+ * @author Dmytro Dashenkov
```
armiol: Let's describe it first before I could look at it.

For now it is completely unclear.

gcd/src/main/java/org/spine3/server/storage/datastore/Kind.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.google.protobuf.Descriptors.Descriptor;
+import com.google.protobuf.Message;
+import org.spine3.type.TypeName;
+import org.spine3.type.TypeUrl;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * @author Dmytro Dashenkov
```
armiol: Same as above. 

I would also consider hiding these classes from `public`. Please review and determine which of those are really `public`.

gcd/src/main/java/org/spine3/server/storage/datastore/DsProperties.java
```diff
@@ -162,7 +162,7 @@ static void makeEventFieldProperties(Event event,
         builder.set(producer_id.toString(), producerId);
     }
 
-    static Predicate<Entity> activeEntityPredicate() {
+    static Predicate<Entity> activedEntityPredicate() {
```
armiol: Something is wrong with English in this name.

And also, why is it in the `DsProperties`? Does not seem to be a good place for such a common thing.

armiol (changes requested):
@dmdashenkov take a look at my comments.

gcd/src/main/java/org/spine3/server/storage/datastore/DsProperties.java
dmdashenkov: The point is that this predicate matches the Datastore `Entity`, not any of our types.
As for me, this should stay in `DsProperties` as it's relevant to the Datastore properties.

BTW, let's discuss if we should remove this at all and migrate to using `Query` instead.

dmdashenkov:
@armiol, PTAL again.

armiol (changes requested):
@dmdashenkov please create `KindShould` as then LGTM.

armiol:
@dmdashenkov one more thing. Could you please append the PR description with the details on how the underlying storage is going to be organized for those, who use `RecordStorage`. A few sentences with an example (say, a couple of `ProjectionRepository` classes) is enough.

dmdashenkov:
@armiol, PTAL again at the changes and the PR description.
Is that what you expected?

armiol (approved):
@dmdashenkov LGTM.

PR #30, opened 2017-04-07
Right margin wrapping

alexander-aleksandrov:
We need to wrap all code to 100 column width.


alexander-aleksandrov:
@dmdashenkov PTAL

gcd/src/test/java/org/spine3/server/storage/datastore/DsAggregateStorageShould.java
```diff
@@ -65,14 +65,16 @@ public static void tearDownClass() {
         datastoreFactory.tearDown();
     }
 
-    @SuppressWarnings("ConstantConditions") // passing null because this parameter isn't used in this implementation
+    @SuppressWarnings("ConstantConditions")
+    // passing null because this parameter isn't used in this implementation
```
dmdashenkov: Please have a tabulation (4 spaces) at the start of the line. This is required to make it obvious that the comment belongs to this specific suppression.

gcd/src/test/java/org/spine3/server/storage/datastore/DsAggregateStorageStatusHandlingShould.java
```diff
@@ -33,7 +33,8 @@
 /**
  * @author Dmytro Dashenkov.
  */
-public class DsAggregateStorageStatusHandlingShould extends AggregateStorageVisibilityHandlingShould {
+public class DsAggregateStorageStatusHandlingShould
+        extends AggregateStorageVisibilityHandlingShould {
```
dmdashenkov: Please make `class` and `extends` keywords start at the same left margin.

dmdashenkov (approved):
@alexander-aleksandrov, LGTM except for two minor issues.

PR #27, opened 2017-03-30
Entity `Columns` support

dmdashenkov:
- Added basic support for the `Columns` storage feature
 - Added such `ColumnType`s:
   - `String`
   - `Integer`
   - `Long`
   - `Boolean`
   - `Timestamp` --> `DateTime`
   - `Message` --> JSON format string
   - `Version` --> `int` version number

PR #31, opened 2017-04-10
Multitenancy support

dmdashenkov:
Added support for _multitenant_ storage structure.
The feature is implemented on the base of the Datastore-native mechanism of _namespaces_.
For more info about the namespaces see [the official doc](https://cloud.google.com/datastore/docs/concepts/multitenancy).

### DIFF
--- a/gcd/src/main/java/org/spine3/server/storage/datastore/dsnative/MultitenantNamespaceSupplier.java
+++ b/gcd/src/main/java/org/spine3/server/storage/datastore/dsnative/MultitenantNamespaceSupplier.java
@@ -29,10 +29,17 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static java.lang.String.format;
 
 /**
+ * A {@link NamespaceSupplier} for multitenant storage factories.
+ *
  * @author Dmytro Dashenkov
  */
 final class MultitenantNamespaceSupplier extends NamespaceSupplier {
 
+    /**
+     * {@inheritDoc}
+     *
+     * @return the {@link Namespace} representing the current tenant {@link TenantId ID}
+     */
     @Override
     public Namespace getNamespace() {
         final TenantIdRetriever retriever = TenantIdRetriever.instance();
@@ -43,6 +50,9 @@ final class MultitenantNamespaceSupplier extends NamespaceSupplier {
         return result;
     }
 
+    /**
+     * @return a human-friendly string representation of the {@link TenantId}
+     */
     private static String tenantIdToSignificantString(TenantId id) {
         final TenantId.KindCase kindCase = id.getKindCase();
         switch (kindCase) {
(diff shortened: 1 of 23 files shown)


dmdashenkov:
@armiol, PTAL.
Please _note_: this PR is targeting the __Entity Columns__ (#27) branch, not `master`.

gcd/src/main/java/org/spine3/server/storage/datastore/dsnative/MultitenantNamespaceSupplier.java
```diff
@@ -0,0 +1,104 @@
...
+        checkNotNull(tenantId);
+        final String stringTenantId = tenantIdToSignificantString(tenantId);
+        final Namespace result = Namespace.of(stringTenantId);
+        return result;
+    }
+
+    /**
+     * @return a human-friendly string representation of the {@link TenantId}
+     */
+    private static String tenantIdToSignificantString(TenantId id) {
```
armiol: This logic should go to `Namespace.of(TenantId)`.

I also do not understand, why you force a user to set the `TenantId.KindCase`. It shouldn't matter at all. Let's discuss the intention.

gcd/src/main/java/org/spine3/server/storage/datastore/dsnative/NamespaceSupplier.java
```diff
@@ -0,0 +1,81 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A factory for the {@linkplain Namespace namespaces}, based on the current multitenancy configuration and
+ * {@linkplain org.spine3.users.TenantId tenant ID}.
+ *
+ * @author Dmytro Dashenkov
+ */
+public abstract class NamespaceSupplier {
```
armiol: First of all, if it is a `Supplier`, then you should use a `Supplier` interface. 

Second of all, the Javadoc should say `A supplier for the ....`.

Also, I don't really understand, why this is a `public` class. Do we expect someone like `@SPI`-user to extend it? Let's discuss.

armiol (changes requested):
@dmdashenkov let's discuss the comments and the general design of the solution.

gcd/src/main/java/org/spine3/server/storage/datastore/dsnative/NamespaceSupplier.java
dmdashenkov: Since we've made the `DatastoreWrapper` `public` and overridable, we need to disclosure `NamespaceSupplier` as well, since `DatastoreWrapper` receives it in the constructor.
Also, one can think of extending this class as about a feature, allowing additional configurations, groupings, etc. for the records in the DB.

dmdashenkov: Also, now I see that we also need to make `Namespace.of` `public` if we do want to disclosure `NamespaceSupplier`.

gcd/src/main/java/org/spine3/server/storage/datastore/dsnative/MultitenantNamespaceSupplier.java
dmdashenkov: `TenantId.KindCase` value is generated automatically when the user sets a value to the `oneof` field `kind` in `TenantId`.
Here I get this value to know where to get the string from.

dmdashenkov:
@armiol, please see my responses.

dmdashenkov:
@armiol, PTAL.

PR #33, opened 2017-04-12
Multitenancy support

dmdashenkov:
Added support for multitenant storage structure.
The feature is implemented on the base of the Datastore-native mechanism of namespaces.
For more info about the namespaces see the official doc.

This PR is a redirection of the PR #31 into `master`.


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@armiol, PTAL.

gcd/src/main/java/org/spine3/server/storage/datastore/DatastoreStorageFactory.java
```diff
@@ -213,15 +226,19 @@ private Builder() {
          * @return self for method chaining
          */
         public Builder setDatastore(Datastore datastore) {
-            this.datastore = checkNotNull(datastore);
+            checkHasNoNamespace(datastore);
```
armiol: This should only apply in case the factory is multitenant. So I would move that to `build()`. And bring back `checkNotNull(datastore)`.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/DsNamespaceSupplierFactory.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private DsNamespaceSupplierFactory() {
+        // Avoid initializing of a utility class
+    }
+
+    /**
+     * Produces an instance of {@link Supplier Supplier&lt;Namespace&gt;} for the given
+     * {@link DatastoreStorageFactory}.
+     *
+     * @param factory the factory to poduce the supplier for
```
armiol: `produce`.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/DsNamespaceValidator.java
```diff
@@ -0,0 +1,65 @@
...
+     * @param datastore the {@link Datastore} to validate the {@linkplain Namespace namespaces} upon
+     */
+    public DsNamespaceValidator(Datastore datastore) {
+        this.namespaceAccess = new NamespaceAccess(datastore);
+    }
+
+    /**
+     * Validates the given {@link Namespace} to match these constraints:
+     * <ul>
+     *     <li>Be not-{@code null}
```
armiol: `be non-{@code null};`,
`be {@linkplain ...}  .`

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/NamespaceAccess.java
```diff
@@ -0,0 +1,160 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Strings.isNullOrEmpty;
+
+/**
+ * A DAO for the Datastore {@link Namespace Namespaces}.
+ *
+ * @author Dmytro Dashenkov
+ */
+class NamespaceAccess implements TenantIndex {
```
armiol: It's `NamespaceIndex`.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/NamespaceAccess.java
```diff
@@ -0,0 +1,160 @@
...
+        Iterators.addAll(newNamespaces, extractedNamespaces);
+
+        // Never delete tenants, only add new ones
+        cache.addAll(newNamespaces);
+    }
+
+    /**
+     * A function retrieving a {@link Namespace} from a given {@link Key}.
+     */
+    private static class NamespaceUnpacker implements Function<Key, Namespace> {
```
armiol: I think it's a factory method of `Namespace`, not a nested class.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/Namespace.java
```diff
@@ -0,0 +1,222 @@
...
+
+    @Override
+    public int hashCode() {
+        return Objects.hashCode(getValue());
+    }
+
+    /**
+     * The converter of the {@code Namespace} into a {@link TenantId} of the specific type..
+     */
+    private enum TenantIdConverter implements Function<Namespace, TenantId> {
```
armiol: It's `TenantIdFunction`, not a converter, as it doesn't implement or extend `Converter`.

You could have a `Converter`, if you merge your `TenantId` -> `Namespace` and `Namespace` -> `TenantId` conversions in an instance of [`Converter`](https://google.github.io/guava/releases/20.0/api/docs/com/google/common/base/Converter.html).

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/NamespaceAccess.java
```diff
@@ -0,0 +1,160 @@
...
+    public void keep(TenantId id) {
+        checkNotNull(id);
+        cache.add(Namespace.of(id));
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Set<TenantId> getAll() {
```
armiol: There are potential concurrency issues with `fetchNamespaces()`, `exists()` and `getAll()`.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/NamespaceSupplier.java
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * Generates a {@link Namespace} based on the current {@linkplain TenantId tenant ID}.
+     *
+     * @return an instance of {@link Namespace} representing either the current tenant ID or an empty string if
+     * the {@linkplain DatastoreStorageFactory storage factory} passed upon the initialization is configured to be
+     * single tenant
+     */
+    @Override
+    public abstract Namespace get();
```
armiol: Please address the warnings.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/Namespace.java
```diff
@@ -0,0 +1,222 @@
...
+        /**
+         * {@inheritDoc}
+         *
+         * <p>Overridden for the nullability contract reset: is {@code non-null} and does not
+         * accept {@code null}s.
+         */
+        @SuppressWarnings("NullableProblems")
+        // Does not accept nulls
+        @Override
+        public abstract TenantId apply(Namespace namespace);
```
armiol: Still a warning to address.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/NamespaceSupplier.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private enum Singleton {
+        INSTANCE;
+        @SuppressWarnings("NonSerializableFieldInSerializableClass")
+        private final NamespaceSupplier singleTenant = new SingleTenantNamespaceSupplier();
+        @SuppressWarnings("NonSerializableFieldInSerializableClass")
```
armiol: Please allow an empty line.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/NamespaceSupplier.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @VisibleForTesting
+    static NamespaceSupplier multitenant() {
+        return Singleton.INSTANCE.multipleTenant;
+    }
+
+    /**
+     * Generates a {@link Namespace} based on the current {@linkplain TenantId tenant ID}.
+     *
+     * @return an instance of {@link Namespace} representing either the current tenant ID or an empty string if
```
armiol: In case of a single tenant, we should probably return the GCD namespace configured by the user — in case it was present at the configuration stage. See the related comment in the `DatastoreStorageFactory.Builder`.

armiol (changes requested):
@dmdashenkov the idea LGTM. 

Please see my comments, especially those on supporting the namespace configured by the user in scope of single-tenant app.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/Namespace.java
dmdashenkov: Can it be `TenantIdTransformer`?

gcd/src/main/java/org/spine3/server/storage/datastore/DatastoreStorageFactory.java
```diff
@@ -251,10 +274,27 @@ public Builder setTypeRegistry(ColumnTypeRegistry<? extends DatastoreColumnType<
...
             }
             return new DatastoreStorageFactory(this);
         }
+
+        private static void checkHasNoNamespace(Datastore datastore) {
+            checkNotNull(datastore);
+            final DatastoreOptions options = datastore.getOptions();
+            final String namespace = options.getNamespace();
+            checkArgument(isNullOrEmpty(namespace),
+                          "Datastore namespace should not be set explicitly.");
```
armiol: Please mention that this is only in case of enabled multitenancy.

gcd/src/main/java/org/spine3/server/storage/datastore/DatastoreStorageFactory.java
```diff
@@ -251,10 +274,27 @@ public Builder setTypeRegistry(ColumnTypeRegistry<? extends DatastoreColumnType<
          */
         public DatastoreStorageFactory build() {
             checkNotNull(datastore);
+            final String defaultNamespace;
```
armiol: Please extract a code portion into the `createNamespaceSupplier` method.

gcd/src/main/java/org/spine3/server/storage/datastore/DatastoreStorageFactory.java
```diff
@@ -218,10 +237,13 @@ public Builder setDatastore(Datastore datastore) {
         }
 
         /**
-         * Switches the {@code DatastoreStorageFactory} that is being built to be or not to be multitenant.
+         * Switches the {@code DatastoreStorageFactory} that is being built to be or not
```
armiol:
```
Enables the multitenancy support for the {@code DatastoreStorageFactory}.

<p>By default this option is {@code false}.

<p> <here you mention that with the multitenancy enabled, the `datastore` passed must not have a namespace set>
```

gcd/src/main/java/org/spine3/server/storage/datastore/DatastoreWrapper.java
```diff
@@ -71,26 +76,41 @@
...
 
-    protected DatastoreWrapper(Datastore datastore) {
-        this.datastore = datastore;
+    private DsNamespaceValidator namespaceValidator;
+
+    /**
+     * Creates a new instance of {@code DatastoreWrapper}.
+     *
+     * @param datastore         {@link Datastore} to wrap
+     * @param namespaceSupplier an instance of {@link Supplier Supplier&lt;Namespace&gt;} to get the
```
armiol:
```
the instance of {@link Supplier Namespace Supplier}, providing the namespaces for Datastore queries.
```

gcd/src/main/java/org/spine3/server/storage/datastore/DsRecordStorage.java
```diff
@@ -364,6 +372,77 @@ protected Kind getKind() {
...
+        /**
+         * @param datastore the {@link DatastoreWrapper} to use in this storage
+         */
+        public Builder<I> setDatastore(DatastoreWrapper datastore) {
+            this.datastore = checkNotNull(datastore);
+            return this;
+        }
+
+        /**
+         * @param multitenant {@code true} if the build storage should be
```
armiol: `if the storage should be`.

gcd/src/main/java/org/spine3/server/storage/datastore/DsRecordStorage.java
```diff
@@ -364,6 +372,77 @@ protected Kind getKind() {
...
+     */
+    public static class Builder<I> {
+
+        private Descriptor descriptor;
+        private DatastoreWrapper datastore;
+        private boolean multitenant;
+        private ColumnTypeRegistry<? extends DatastoreColumnType<?, ?>> columnTypeRegistry;
+
+        private Builder() {
+            // Avoid direct initialization
```
armiol: `Prevent`

gcd/src/main/java/org/spine3/server/storage/datastore/DsRecordStorage.java
```diff
@@ -364,6 +372,77 @@ protected Kind getKind() {
...
+        public Builder<I> setDatastore(DatastoreWrapper datastore) {
+            this.datastore = checkNotNull(datastore);
+            return this;
+        }
+
+        /**
+         * @param multitenant {@code true} if the build storage should be
+         * {@link org.spine3.server.storage.Storage#isMultitenant multitenant} or not
+         */
+        public Builder<I> setMultitenant(boolean multitenant) {
```
armiol: We need to be careful with this. It should not have the value different from the storage factory.

I can see this method (and the whole `Builder`) is public, so you might as well try to hide it. And maybe the whole `DsRecordStorage` should be removed from `public`, I am not sure. Please investigate.

gcd/src/main/java/org/spine3/server/storage/datastore/Kind.java
```diff
@@ -40,12 +42,25 @@
             "Datastore kind cannot start with \"__\". See https://cloud.google.com/datastore/docs/concepts/entities#kinds_and_identifiers for more info.";
```
armiol: Please split the line to fit 100 columns wide.

gcd/src/main/java/org/spine3/server/storage/datastore/Kind.java
```diff
@@ -66,6 +81,14 @@ public static Kind of(TypeName typeName) {
         return new Kind(typeName.value());
     }
 
+    /**
+     * Produces a {@code Kind} representing the datastore namespace kind.
+     */
+    @Internal
+    public static Kind ofNamespace() {
```
armiol: It's weird that this is `public`. Not a big deal for now, as it's internal. However, it may be a sign that the package is chosen incorrectly.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/DatastoreTenants.java
```diff
@@ -0,0 +1,79 @@
...
+     * when creating an instance of {@link org.spine3.server.BoundedContext BoundedContext}.
+     *
+     * <p>An example of creating a multitenant
+     * {@link org.spine3.server.BoundedContext BoundedContext} using the Datastore Storage:
+     * <pre>
+     *     <code>
+     *         final Datastore myDatastoreConfig = myDatastoreOptions.getService();
+     *
+     *         // Create DatastoreStorageFactories using this instance of Datastore
+     *         final Supplier{@literal <}StorageFactory{@literal >} dsStorageFactorySupplier = getDatastoreStorageFactorySupplier(myDatastoreConfig);
```
armiol: Please try to fit 100 columns wide.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/DatastoreTenants.java
```diff
@@ -0,0 +1,79 @@
...
+     *
+     *         // Use the same instance for the TenantIndex
+     *         final TenantIndex myTenantIndex = DatastoreTenants.index(myDatastoreConfig);
+     *
+     *         // Pass both to the BoundedContext.Builder
+     *         final BoundedContext multitenantAppBc = BoundedContext.newBuilder()
+     *                                                               .setStorageFactorySupplier(dsStorageFactorySupplier)
+     *                                                               .setTenantIndex(myTenantIndex)
+     *                                                               // set other params
```
armiol: I think you should add `.setMultitenant(true)` to be consistent.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/NamespaceIndex.java
```diff
@@ -0,0 +1,166 @@
...
+     * Checks if the Datastore has the given {@linkplain Namespace}, i.e. there is at least one
+     * {@linkplain com.google.cloud.datastore.Entity Entity} in this {@linkplain Namespace}.
+     *
+     * @param namespace the {@linkplain Namespace} yo look for
+     * @return {@code true} if there is at least one
+     * {@linkplain com.google.cloud.datastore.Entity Entity} in this {@linkplain Namespace} or the corresponding
+     * {@link TenantId} has been put into the index by a call to {@link #keep(TenantId)},
+     * {@code false} otherwise
+     */
+    boolean exists(Namespace namespace) {
```
armiol: `contains`.

Otherwise `NamespaceIndex.exists(namespace)` reads improperly.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/NamespaceSupplier.java
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @see org.spine3.server.storage.StorageFactory#isMultitenant
+     */
+    public static NamespaceSupplier instance(boolean multitenant, @Nullable String defaultNamespace) {
+        if (multitenant) {
+            return multitenant();
+        } else {
+            final String namespace = defaultNamespace != null
+                                     ? defaultNamespace
+                                     : "";
```
armiol: You have the same String literal here and in the method below. Please extract that.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/SingleTenantNamespaceSupplier.java
```diff
@@ -0,0 +1,57 @@
...
+package org.spine3.server.storage.datastore.tenant;
+
+/**
+ * A {@link NamespaceSupplier} for single-tenant storage factories.
+ *
+ * @author Dmytro Dashenkov
+ */
+final class SingleTenantNamespaceSupplier extends NamespaceSupplier {
+
+    private static final String DEFAULT_NAMESPACE = "";
```
armiol, line 34: Here it is for the third time. See my comments above. Something isn't right.

I'd say that this class only should be a keeper of the single-tenant logic for namespace suppliers.

gcd/src/main/java/org/spine3/server/storage/datastore/type/DatastoreTypeRegistryFactory.java
```diff
@@ -56,7 +56,7 @@ private DatastoreTypeRegistryFactory() {
     }
 
     /**
-     * Retrieves a default {@link ColumnTypeRegistry ColumnTypeRegistry{@literal <}? extends DatastoreColumnType>}
+     * Retrieves a default {@link ColumnTypeRegistry ColumnTypeRegistry&lt;? extends DatastoreColumnType&gt;}
```
armiol: The description repeats the signature. 

Instead of having fun with `&lt;` and `&gt;` (and not only here; every time repeating what's written in code), please try to make the aim of the method more clear.

gcd/src/main/java/org/spine3/server/storage/datastore/DsRecordStorage.java
```diff
@@ -364,6 +372,77 @@ protected Kind getKind() {
...
+        private ColumnTypeRegistry<? extends DatastoreColumnType<?, ?>> columnTypeRegistry;
+
+        private Builder() {
+            // Avoid direct initialization
+        }
+
+        /**
+         * @param descriptor the record state {@link Descriptor}
+         */
+        public Builder<I> setDescriptor(Descriptor descriptor) {
```
armiol: It's a weird mandatory parameter for the public API. Please try to find some way to submit the same information easier; for instance, by requesting the `Message` class instead.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/NamespaceIndex.java
```diff
@@ -0,0 +1,166 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A DAO for the Datastore {@link Namespace Namespaces}.
+ *
+ * @author Dmytro Dashenkov
+ */
+@ThreadSafe
+class NamespaceIndex implements TenantIndex {
```
armiol, line 49: I think we should try to test this class. At the moment it's not easy, as the Datastore query is hard-coded, so it's impossible to mock. 

I'd suggest to extract the query into a separate internal class. Then you'll be able to prove this class is thread-safe indeed in tests.

armiol (changes requested):
@dmdashenkov please see my comments.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/SingleTenantNamespaceSupplier.java
dmdashenkov: Done

gcd/src/main/java/org/spine3/server/storage/datastore/Kind.java
dmdashenkov: This is better than making the ancillary kind constructor `public` (or any but `private`).

I agree that semantically this may belong to `tenant` package, but this is an exceptional case when we walk around the kind validation rules. Therefore, I believe, this is the best way to prevent the user from breaking through the validation by himself.

dmdashenkov:
@armiol, PTAL again.

gcd/src/main/java/org/spine3/server/storage/datastore/Kind.java
```diff
@@ -66,6 +83,14 @@ public static Kind of(TypeName typeName) {
         return new Kind(typeName.value());
     }
 
+    /**
+     * Produces a {@code Kind} representing the datastore namespace kind.
```
armiol: `Datastore` should be capitalized.

Could you please check the changes for this issue?

gcd/src/main/java/org/spine3/server/storage/datastore/ProjectId.java
```diff
@@ -0,0 +1,98 @@
...
+import com.google.common.base.Objects;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A value-type representing the GAE project ID.
+ *
+ * @author Dmytro Dashenkov
+ */
+public class ProjectId {
```
armiol: Looks `final`.

gcd/src/main/java/org/spine3/server/storage/datastore/tenant/DatastoreTenants.java
```diff
@@ -0,0 +1,82 @@
...
+import org.spine3.server.tenant.TenantIndex;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A factory of the Datastore-specific Tenant related objects.
+ *
+ * @author Dmytro Dashenkov
+ */
+public class DatastoreTenants {
```
armiol: As discussed vocally, this should be made `final`.

The same applies to all the utility classes like this.

armiol (approved):
@dmdashenkov LGTM except for a few comments. Please resolve those before merging.

Also, as discussed vocally, let's create an issue in `gae-java` to simplify the creation of `BoundedContext` based upon GAE Datastore. The idea is to require less parameters from the framework user (e.g. `Datastore` instance and `multitenancy` flag).

PR #38, opened 2017-04-27
Migrate to Spine core of version 0.8.35-SNAPSHOT

dmdashenkov:
Added support for the Spine core v0.8.35-SNAPSHOT repackaging.

This is required to make is possible to use the GAE Storage with Spine core of that version.


dmdashenkov:
@armiol, PTAL.

PR #48, opened 2017-06-09
AggregateEventRecord ordering fix

dmdashenkov:
As we faced the issue with the event timestamp overlapping, we need to sort the AggregateEventRecords also by version.

To make such ordering available on the Datastore, the framework used should add some indexes for the types.


dmdashenkov:
@armiol, PTAL.
Note that the PR is still in the WIP status, though no major changes may be required. Most probably, the biggest change will be applied to the `./README.MD` and `./gcd/config/indexes.yaml` files, but not to the actual code.

gcd/src/main/java/io/spine/server/storage/datastore/DsProperties.java
```diff
@@ -38,6 +47,14 @@
 @SuppressWarnings("UtilityClass")
 class DsProperties {
 
+    private static final String CREATED_PROPERTY = "created";
```
armiol: These are the members of one enum...

gcd/src/main/java/io/spine/server/storage/datastore/DsProperties.java
```diff
@@ -38,6 +47,14 @@
...
+    private static final String VERSION_PROPERTY = "version";
+    private static final String SNAPSHOT_PROPERTY = "snapshot";
+
+    private static final OrderBy BY_CREATED_PROPERTY = desc(CREATED_PROPERTY);
```
armiol: ...and these are the members of another.

armiol (dismissed):
@dmdashenkov LGTM in general. Please see my comments. Also, there are some portions of the commented code, which must be removed.

dmdashenkov:
@alexander-yevsyukov, PTAL.

README.md
```diff
@@ -18,6 +18,9 @@ be set.
 
 The configuration file is located at `./gcd/config/index.yaml`. 
 
+Please note, that it is required to add the specified in the `./gcd/config/index.yaml` indexes for
```
alexander-yevsyukov: The `specified` what?

gcd/config/index.yaml
```diff
@@ -1,6 +1,25 @@
...
     - name: created
+
+  # Indexes for the Aggregates.
+  #
+  # Each custom Aggregate type must have such indexes in order to be retrieved properly.
+  #
+  # The Datastore kind represents the Protobuf type name of the Aggregate state. Both the property
+  # names and the index directions should be kept unchanged.
+
+  - kind: spine.test.aggregate.Project
```
alexander-yevsyukov: If this is a sample of how an index entry should look like, why isn't it a part of a comment block?

I would suggest re-naming the kind to be ending with `MyAggregate`, and the package being `myapp.example` or something like that. We need to highlight the fact that it doesn't belong to Spine.

alexander-yevsyukov (changes requested):
Please see my comments and questions.

dmdashenkov:
@alexander-yevsyukov, PTAL again.

gcd/config/index.yaml
```diff
@@ -1,3 +1,5 @@
+# THe GAE Datastore sample indexes.
```
alexander-yevsyukov: Please fix the spelling.

gcd/config/index.yaml
```diff
@@ -15,7 +17,16 @@ indexes:
   # The Datastore kind represents the Protobuf type name of the Aggregate state. Both the property
   # names and the index directions should be kept unchanged.
 
-  - kind: spine.test.aggregate.Project
+  - kind: myapp.example.MyFirstAggregate
```
alexander-yevsyukov: Please have a comment following this line that would encourage the user to replace only this line with an aggregate type name. And only it.

gcd/config/index.yaml
alexander-yevsyukov: Please rename the folder to `datastore`.

alexander-yevsyukov (changes requested):
Please see my comments.

dmdashenkov:
@alexander-yevsyukov, PTAL again.

datastore/src/main/java/io/spine/server/storage/datastore/DsProperties.java
```diff
@@ -49,6 +59,22 @@ static void addAggregateIdProperty(Object aggregateId, Entity.Builder entity) {
         entity.set(aggregate_id.toString(), propertyValue);
     }
 
+    static void addCreatedProperty(Timestamp when, Entity.Builder entity) {
+        final long millis = Timestamps.toMillis(when);
+        final Date date = new Date(millis);
+        final DateTime dateTime = DateTime.copyFrom(date);
+        entity.set(AggregateEventRecordProperty.CREATED.toString(), dateTime);
```
alexander-yevsyukov: Please have static import for `AggregateEventRecordProperty` values. It would be easier to read.

datastore/src/main/java/io/spine/server/storage/datastore/DsProperties.java
```diff
@@ -67,10 +93,51 @@ static boolean isDeleted(Entity entity) {
...
+
+        @Override
+        public String toString() {
+            return name().toLowerCase();
+        }
+    }
+
+    private enum AggregateEventRecordOrdering {
+
+        BY_CREATED_PROPERTY(desc(AggregateEventRecordProperty.CREATED.toString())),
```
alexander-yevsyukov: The `PROPERTY` suffix can be omitted. `BY_` is quite enough.

datastore/src/main/java/io/spine/server/storage/datastore/DsProperties.java
```diff
@@ -67,10 +93,51 @@ static boolean isDeleted(Entity entity) {
...
+    }
+
     private static boolean hasFlag(Entity entity, String flagName) {
         final boolean result =
                 entity.contains(flagName)
                         && entity.getBoolean(flagName);
         return result;
     }
+
+    private enum AggregateEventRecordProperty {
```
alexander-yevsyukov: Why don't we have a common interface for fields and implement it here? This way it would be easier to find fields across the codebase.

alexander-yevsyukov: In Java 8 we would automatically get lowercasing by having default `toString()` impl.

alexander-yevsyukov (changes requested):
Please see my comments.

datastore/src/main/java/io/spine/server/storage/datastore/DsProperties.java
dmdashenkov: Unfortunately, we cannot `import static` members of a private type in Java.

dmdashenkov:
@alexander-yevsyukov, PTAL again.

datastore/src/main/java/io/spine/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -124,24 +140,38 @@ public void writeEventCountAfterLastSnapshot(I id, int eventCount) {
...
+                                .getContext()
+                                .getVersion();
+                break;
+            case SNAPSHOT:
+                recordId = SNAPSHOT + stringId;
+                version = record.getSnapshot()
+                                .getVersion();
+                break;
+            default:
+                throw new IllegalArgumentException(record.getKindCase()
```
alexander-yevsyukov: Arguably, it looks like a case for `IllegalStateException` because _this_ part of code cannot handle an `AggregateEventRecord`. Anyway, we need a message which explains the problem in more details than just a name of the kind case.

datastore/src/main/java/io/spine/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -198,15 +223,16 @@ public int compare(AggregateEventRecord o1, AggregateEventRecord o2) {
...
 
     /**
-     * Generates an identifier of the Datastore record basing on the given {@code Aggregate} identifier.
+     * Generates an identifier of the Datastore record basing on the given {@code Aggregate}
+     * identifier.
      *
      * @param id an identifier of the {@code Aggregate}
      * @return the Datastore record ID
      */
     protected DatastoreRecordId generateDatastoreId(I id) {
```
alexander-yevsyukov: It looks like a conversion method. We do not really generate inside. We convert. A better name would be `toDatastoreId()` or `toRecordId()`.

Assuming we are under `datastore` package, the returned type can be also named `RecordId`.

datastore/src/main/java/io/spine/server/storage/datastore/DsProperties.java
```diff
@@ -49,6 +60,22 @@ static void addAggregateIdProperty(Object aggregateId, Entity.Builder entity) {
         entity.set(aggregate_id.toString(), propertyValue);
     }
 
+    static void addCreatedProperty(Timestamp when, Entity.Builder entity) {
```
alexander-yevsyukov: Have the `Builder` parameter first.

alexander-yevsyukov: The `Property` suffix can be omitted. We have it all under the `DsProperties` class.

datastore/src/main/java/io/spine/server/storage/datastore/DsProperties.java
```diff
@@ -49,6 +60,22 @@ static void addAggregateIdProperty(Object aggregateId, Entity.Builder entity) {
...
+        final DateTime dateTime = DateTime.copyFrom(date);
+        entity.set(AggregateEventRecordProperty.created.toString(), dateTime);
+    }
+
+    static void addVersionProperty(Version version, Entity.Builder entity) {
+        final int number = version.getNumber();
+        entity.set(AggregateEventRecordProperty.version.toString(), number);
+    }
+
+    static void markSnapshotProperty(boolean snapshot, Entity.Builder entity) {
```
alexander-yevsyukov: How about naming it `markAsSnapshot()`?

datastore/src/main/java/io/spine/server/storage/datastore/DsProperties.java
```diff
@@ -49,6 +60,22 @@ static void addAggregateIdProperty(Object aggregateId, Entity.Builder entity) {
         entity.set(aggregate_id.toString(), propertyValue);
     }
 
+    static void addCreatedProperty(Timestamp when, Entity.Builder entity) {
+        final long millis = Timestamps.toMillis(when);
+        final Date date = new Date(millis);
+        final DateTime dateTime = DateTime.copyFrom(date);
+        entity.set(AggregateEventRecordProperty.created.toString(), dateTime);
```
alexander-yevsyukov: We need to be consistent in naming enums. Some of them are lowercase, some of them are uppercase. I think it would be better to have them uppercase with proper `toString()` and static import of `AggregateEventRecordProperty`.  

You can also have a static method that would accept 1) `Entity.Builder` 2) property instance (as an interface) and 3) property value (as a generified argument). And the method would call `toString()`.

datastore/src/main/java/io/spine/server/storage/datastore/DsProperties.java
```diff
@@ -67,10 +94,58 @@ static boolean isDeleted(Entity entity) {
...
                         && entity.getBoolean(flagName);
         return result;
     }
+
+    private enum AggregateEventRecordProperty implements StorageField {
+
+        /**
+         * A property storing the Event creation time.
+         */
+        created,
```
alexander-yevsyukov: As commented above, let's standardize on uppercase and `toString()` that would convert to lowercase. In Java 8 it would be default method.

alexander-yevsyukov (changes requested):
Please see my comments.

datastore/src/main/java/io/spine/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -69,15 +81,18 @@
...
      *
-     * The aggregate snapshots are stored under an ID composed from {@code SNAPSHOT} and the aggregate ID.
+     * <p>The aggregate snapshots are stored under an ID composed from {@code SNAPSHOT} and
+     * the aggregate ID.
      */
     private static final String SNAPSHOT = "SNAPSHOT";
 
-    private static final TypeName AGGREGATE_LIFECYCLE_KIND = TypeName.from(LifecycleFlags.getDescriptor());
-    private static final TypeUrl AGGREGATE_RECORD_TYPE_URL = TypeUrl.from(AggregateEventRecord.getDescriptor());
+    private static final TypeName AGGREGATE_LIFECYCLE_KIND = TypeName.from(
```
alexander-yevsyukov: Consider new line after `=`.

datastore/src/main/java/io/spine/server/storage/datastore/DsProperties.java
alexander-yevsyukov: Or, have an abstract method for `AggregateEventRecordProperty` and set values in impl.

datastore/src/main/java/io/spine/server/storage/datastore/DsProperties.java
```diff
@@ -67,10 +94,58 @@ static boolean isDeleted(Entity entity) {
...
+    static OrderBy byVersion() {
+        return AggregateEventRecordOrdering.BY_VERSION.getOrdering();
+    }
+
+    static OrderBy byRecordType() {
+        return AggregateEventRecordOrdering.BY_SNAPSHOT.getOrdering();
+    }
+
     private static boolean hasFlag(Entity entity, String flagName) {
         final boolean result =
```
alexander-yevsyukov: Please fix wrapping here.

datastore/src/main/java/io/spine/server/storage/datastore/DsProperties.java
```diff
@@ -67,10 +94,58 @@ static boolean isDeleted(Entity entity) {
...
+        /**
+         * A boolean property storing {@code true} if the Record represents a Snapshot and
+         * {@code false} otherwise.
+         */
+        snapshot
+    }
+
+    private enum AggregateEventRecordOrdering {
+
+        BY_CREATED(desc(AggregateEventRecordProperty.created.toString())),
```
alexander-yevsyukov: Can we make `desc` and `asc` accept `StorageField` instances?

datastore/src/main/java/io/spine/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -162,16 +163,16 @@ protected void writeRecord(I id, AggregateEventRecord record) {
...
+                                                  record.getKindCase());
         }
         final Key key = keyFor(datastore, Kind.of(stateTypeName), of(recordId));
         final Entity incompleteEntity = messageToEntity(record, key);
         final Entity.Builder builder = Entity.newBuilder(incompleteEntity);
-        addAggregateIdProperty(stringId, builder);
-        addCreatedProperty(record.getTimestamp(), builder);
-        addVersionProperty(version, builder);
-        markSnapshotProperty(kind == KindCase.SNAPSHOT, builder);
+        addAggregateIdProperty(builder, stringId);
```
alexander-yevsyukov: Please have `addAggregateId()`, `addCreated()`, `addVersion()`.

dmdashenkov:
@alexander-yevsyukov, PTAL again.

alexander-yevsyukov (changes requested):
Please see my comment.

alexander-yevsyukov (approved):
LGTM

PR #55, opened 2017-09-01
Fix Entity query perormance

dmdashenkov:
Until now we treated the Entity Column query better than the ID query.

Since this PR, the ID query has higher priority than the Entity Column lookup.
If the IDs are specified in the `EntityQuery`, then the Datastore query is performed over the IDs.
If the Column Filers are also present, then the retrieved from the Datastore entities are filtered in memory.

_Iff_ the IDs are absent, then the actual Column-based Datastore query is performed.


dmdashenkov:
@armiol, PTAL.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: java
 
 jdk:
-  - oraclejdk7
+  - oraclejdk8
```
armiol, line 4: How come? I wouldn't change that just in case. Let's discuss.

datastore/src/main/java/io/spine/server/storage/datastore/DsRecordStorage.java
```diff
@@ -444,6 +456,80 @@ EntityRecord getRecordFromEntity(Entity entity) {
...
+        private final ColumnFilterAdapter adapter;
+        private final Iterable<CompositeQueryParameter> queryParams;
+
+        private EntityColumnPredicate(Iterable<CompositeQueryParameter> queryParams,
+                                      ColumnFilterAdapter adapter) {
+            this.adapter = adapter;
+            this.queryParams = queryParams;
+        }
+
+        @SuppressWarnings("EnumSwitchStatementWhichMissesCases") // Only valuable cases covered
```
armiol: You can list them just above `default:` and remove this warning suppression.

datastore/src/test/java/io/spine/server/storage/datastore/DsRecordStorageShould.java
```diff
@@ -276,32 +344,39 @@ protected TestConstCounterEntity(ProjectId id) {
             this.creationTime = getCurrentTime();
         }
 
+        @javax.persistence.Column
```
armiol, line 347: This is going to be changed. If not already.

armiol (changes requested):
@dmdashenkov Mostly LGTM. Not sure I understand everything, let's talk.

Also, we're getting a proper `Column` annotation really soon. Maybe it's even merged in already. I'd wait until then — should be no longer than a day.

PR #56, opened 2017-09-04
Custom column names for storing and better `EntityQuery` performance

dmytro-grankin:
This PR  brings support for custom entity column names for storing.  For the details please see the [PR](https://github.com/SpineEventEngine/core-java/pull/577).

Also the performance for entity queries was improved. Until now we treated the Entity Column query better than the ID query.

Since this PR, the ID query has higher priority than the Entity Column lookup.
If the IDs are specified in the `EntityQuery`, then the Datastore query is performed over the IDs.
If the Column Filters are also present, then the retrieved from the Datastore entities are filtered in memory.

If the IDs are absent, then the actual Column-based Datastore query is performed.


dmytro-grankin:
@armiol PTAL.

armiol (approved):
@dmytro-grankin LGTM.

PR #59, opened 2017-09-19
Allow Namespace autocreation

### LINKED ISSUE #58: Allow Namespace creation
It turns out that It's impossible to create a Datastore namespace with some admin tools. The only way to do that is to write a record into the namespace.

That's why we should allow the users of the library to create the namespaces instead of failing in this case.

dmdashenkov:
From now we allow dynamic creation of the Datastore namespaces. 

To Create a namespace for a Tenant ID, the user of the library mustn't perform any special actions. The namespaces are created by the rules of the Datastore when the first record with such namespace is written to the DB and destroyed when there are no more records associated with that namespace.

Fixes #58.


dmdashenkov:
@dmytro-grankin, PTAL.

dmytro-grankin (approved):
@dmdashenkov LGTM.

PR #61, opened 2017-10-11
Optimize reading of aggregates

### LINKED ISSUE #60: Optimize `AggregateEventRecord` read
Currently, the `AggregateStorage.read(id)` works as follows:
 1. Read all `AggregateEventRecord`s with given Aggregate ID (`.historyBackward()`).
 2. Skip all the records till the last snapshot.
 3. If a snapshot is not found - return all records.
 4. If a snapshot is present, return the snapshot and all the event records of events that occurred after the snapshot was made.

We need to optimize the read to make sure, that if a snapshot is present, we __do not__ read all the records, but only the ones that are required.

dmytro-grankin:
This PR fixes #60.

Recently, `Storage.read(...)` method was reworked. See details in the [PR](https://github.com/SpineEventEngine/core-java/pull/596).

`DsAggregateStorage` was updated accordingly and now it uses a batch size from the newly introduced `AggregateReadRequest` to read a history of an aggregate by batches.

Also, ordering of events in `DsAggregateStorage` was fixed. Before, events were sorted by creation time and only then by version. But ordering by version is more important, so current storage implementation sorts events by version and only then by creation time.

Spine version was increased to `0.9.76-SNAPSHOT`.

### DIFF
--- a/datastore/src/main/java/io/spine/server/storage/datastore/DsAggregateStorage.java
+++ b/datastore/src/main/java/io/spine/server/storage/datastore/DsAggregateStorage.java
@@ -223,8 +223,8 @@ public class DsAggregateStorage<I> extends AggregateStorage<I> {
                     .setKind(stateTypeName.value())
                     .setFilter(eq(aggregate_id.toString(),
                                   idString))
-                    .setOrderBy(byCreatedTime(),
-                                byVersion(),
+                    .setOrderBy(byVersion(),
+                                byCreatedTime(),
                                 byRecordType())
                     .setLimit(limit)
                     .build();
--- a/datastore/src/main/java/io/spine/server/storage/datastore/DsAggregateStorage.java
+++ b/datastore/src/main/java/io/spine/server/storage/datastore/DsAggregateStorage.java
@@ -21,9 +21,11 @@
 package io.spine.server.storage.datastore;
 
 import com.google.cloud.datastore.Entity;
+import com.google.cloud.datastore.EntityQuery;
 import com.google.cloud.datastore.Key;
 import com.google.cloud.datastore.Query;
 import com.google.cloud.datastore.StructuredQuery;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.common.base.Predicate;
@@ -32,6 +34,7 @@ import com.google.protobuf.Message;
 import io.spine.core.Version;
 import io.spine.server.aggregate.AggregateEventRecord;
 import io.spine.server.aggregate.AggregateEventRecord.KindCase;
+import io.spine.server.aggregate.AggregateReadRequest;
 import io.spine.server.aggregate.AggregateStorage;
 import io.spine.server.entity.LifecycleFlags;
 import io.spine.string.Stringifiers;
@@ -150,7 +153,8 @@ public class DsAggregateStorage<I> extends AggregateStorage<I> {
         final KindCase kind = record.getKindCase();
         switch (kind) {
             case EVENT:
-                recordId = Stringifiers.toString(record.getEvent().getId());
+                recordId = Stringifiers.toString(record.getEvent()
+                                                       .getId());
                 version = record.getEvent()
                                 .getContext()
                                 .getVersion();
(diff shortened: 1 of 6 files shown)


dmytro-grankin:
@armiol PTAL.

datastore/src/main/java/io/spine/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -174,19 +178,18 @@ protected void writeRecord(I id, AggregateEventRecord record) {
         datastore.createOrUpdate(builder.build());
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * <p>The resulting iterator will fetch {@linkplain AggregateEventRecord events}
+     * by batches. Size of a batch is specified by the given {@link AggregateReadRequest}.
```
armiol:
```
...will fetch {... events} batch by batch.

<p>The size of the batch is specified ...
```

datastore/src/main/java/io/spine/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -210,6 +213,22 @@ public Key apply(@Nullable Entity input) {
...
+    @VisibleForTesting
+    EntityQuery historyBackwardQuery(AggregateReadRequest<I> request) {
+        checkNotNull(request);
+        final String idString = Stringifiers.toString(request.getRecordId());
+        final int limit = request.getBatchSize();
+        return Query.newEntityQueryBuilder()
+                    .setKind(stateTypeName.value())
+                    .setFilter(eq(aggregate_id.toString(),
+                                  idString))
+                    .setOrderBy(byCreatedTime(),
```
armiol: One more issue should be addressed within this method. 

At the moment we order by created time and only then by version. In fact, ordering by version is more important and must come first. Please address that and include to the PR description.

armiol (commented):
@dmytro-grankin please see my comments. 

Also, as discussed, let's create a test, that verifies reading of an `Aggregate` instance in case snapshot trigger is changed to a smaller value in the middle of operation.

datastore/src/main/java/io/spine/server/storage/datastore/DsAggregateStorage.java
dmytro-grankin: Done.

datastore/src/main/java/io/spine/server/storage/datastore/DsAggregateStorage.java
dmytro-grankin: Done.

dmytro-grankin:
@armiol PTAL.

datastore/src/test/java/io/spine/server/storage/datastore/DsAggregateStorageShould.java
```diff
@@ -98,6 +112,67 @@ public void fail_to_write_invalid_record() {
...
+        final AggregateReadRequest<ProjectId> request = new AggregateReadRequest<>(newId(),
+                                                                                   batchSize);
+        final EntityQuery historyBackwardQuery = storage.historyBackwardQuery(request);
+
+        final int queryLimit = historyBackwardQuery.getLimit();
+        assertEquals(batchSize, queryLimit);
+    }
+
+    @Test
+    public void allow_aggregate_loading_if_snapshot_trigger_was_decreased_in_AggregateRepository() {
```
armiol: To emphasize those parts that matter, I would name it 
`still_load_aggregates_properly_after_snapshot_trigger_decrease_at_runtime`.

datastore/src/test/java/io/spine/server/storage/datastore/DsAggregateStorageShould.java
```diff
@@ -98,6 +112,67 @@ public void fail_to_write_invalid_record() {
...
+
+        final int minimalSnapshotTrigger = 1;
+        repository.setSnapshotTrigger(minimalSnapshotTrigger);
+        final ProjectAggregate aggregate = repository.find(id)
+                                                     .get();
+        assertEquals(tasksCount, aggregate.getState()
+                                          .getTaskCount());
+    }
+
+    private static void register(AggregateRepository repository) {
```
armiol: Let's move things like this one to `Given` (similar to what we do in `core-java`). This method is both excessive in the test code and weird — as it's `static`, looks like a registration, but in fact creates a bounded context each time.

Let's discuss if needed. But at the moment this piece of code is difficult to reuse properly.

armiol (changes requested):
@dmytro-grankin Mostly LGTM. However, a few things still to address.

dmytro-grankin:
@armiol PTAL.

datastore/src/main/java/io/spine/server/storage/datastore/DsStandStorage.java
```diff
@@ -75,7 +75,8 @@ public boolean delete(AggregateStateId id) {
     @Nullable
```
armiol: I think it's not `@Nullable`. As long as you modify this method, please pay attention to things like that.

datastore/src/test/java/io/spine/server/storage/datastore/DsAggregateStorageShould.java
```diff
@@ -98,6 +109,56 @@ public void fail_to_write_invalid_record() {
...
+
+        final int minimalSnapshotTrigger = 1;
+        repository.setSnapshotTrigger(minimalSnapshotTrigger);
+        final ProjectAggregate aggregate = repository.find(id)
+                                                     .get();
+        assertEquals(tasksCount, aggregate.getState()
+                                          .getTaskCount());
+    }
+
+    private static class ProjectAggregateRepository
```
armiol: Please move this class to `Given` (like we do in `core-java`) and have a class-level comment for it.

armiol (approved):
@dmytro-grankin LGTM with a couple of issues to address.

PR #65, opened 2018-03-06
Filter lifecycle flags with a DB query where possible

dmdashenkov:
Before this PR, all the filtering by `archived` and `deleted` fields happened in-memory, thus causing a significant overuse of the Datastore API.

From now, the filtering is performed on the database side where possible (in `ProjectionStorage` in particular).


datastore/src/main/java/io/spine/server/storage/datastore/DsFilters.java
```diff
@@ -87,6 +95,10 @@ private DsFilters() {
         // Prevent utility class fromm being initialized.
     }
 
+    static Filter activeEntity() {
```
armiol: Please document.

datastore/src/main/java/io/spine/server/storage/datastore/DsProjectionStorageDelegate.java
```diff
@@ -0,0 +1,142 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.server.storage.datastore.DsFilters.activeEntity;
+
+/**
+ * A {@link io.spine.server.storage.RecordStorage RecordStorage} to which
+ * {@link DsProjectionStorage} delegates its operations.
+ *
+ * @author Dmytro Dashenkov
+ */
+public class DsProjectionStorageDelegate<I> extends DsRecordStorage<I> {
```
armiol: Please append the documentation with more on why we have it at all. It isn't clear at all. But it should be — to anyone who's going to deal with this class in future.

datastore/src/main/java/io/spine/server/storage/datastore/DsProjectionStorageDelegate.java
```diff
@@ -0,0 +1,142 @@
...
+    protected StructuredQuery<Entity> buildAllQuery(TypeUrl typeUrl) {
+        final String entityKind = kindFrom(typeUrl).getValue();
+        final StructuredQuery<Entity> query = Query.newEntityQueryBuilder()
+                                                   .setKind(entityKind)
+                                                   .setFilter(activeEntity())
+                                                   .build();
+        return query;
+    }
+
+    public static <I> Builder<I> builder() {
```
armiol: We use `newBuilder()` in `core-java`. Can we please use this naming approach here as well?

datastore/src/main/java/io/spine/server/storage/datastore/DsProjectionStorageDelegate.java
```diff
@@ -0,0 +1,142 @@
...
+    public static class Builder<I> {
+
+        private Descriptor descriptor;
+        private DatastoreWrapper datastore;
+        private boolean multitenant;
+        private ColumnTypeRegistry<? extends DatastoreColumnType<?, ?>> columnTypeRegistry;
+        private Class<I> idClass;
+
+        private Builder() {
+            // Prevent direct initialization.
```
armiol: This should be a Javadoc according to the latest Spine codestyle.

datastore/src/main/java/io/spine/server/storage/datastore/DsRecordStorage.java
```diff
@@ -116,11 +117,12 @@ public EntityRecord apply(@Nullable Entity input) {
      * @param descriptor the descriptor of the type of messages to save to the storage
      * @param datastore  the Datastore implementation to use
      */
-    protected DsRecordStorage(Descriptor descriptor,
-                              DatastoreWrapper datastore,
-                              boolean multitenant,
-                              ColumnTypeRegistry<? extends DatastoreColumnType<?, ?>> columnTypeRegistry,
-                              Class<I> idClass) {
+    protected DsRecordStorage(
+            Descriptor descriptor,
```
armiol: I would leave this on the previous line (as before).

datastore/src/main/java/io/spine/server/storage/datastore/DsProjectionStorageDelegate.java
```diff
@@ -0,0 +1,142 @@
...
+         */
+        public Builder<I> setDatastore(DatastoreWrapper datastore) {
+            this.datastore = checkNotNull(datastore);
+            return this;
+        }
+
+        /**
+         * @param multitenant {@code true} if the storage should be
+         *                    {@link io.spine.server.storage.Storage#isMultitenant multitenant}
+         *                    or not
```
armiol: `,` at the end of the previous line `{@code false} otherwise` should go instead of this line. Right now it makes not much sense.

datastore/src/main/java/io/spine/server/storage/datastore/DsProjectionStorageDelegate.java
```diff
@@ -0,0 +1,142 @@
...
+        private DatastoreWrapper datastore;
+        private boolean multitenant;
+        private ColumnTypeRegistry<? extends DatastoreColumnType<?, ?>> columnTypeRegistry;
+        private Class<I> idClass;
+
+        private Builder() {
+            // Prevent direct initialization.
+        }
+
+        public Builder<I> setStateType(TypeUrl stateTypeUrl) {
```
armiol: Please document.

datastore/src/main/java/io/spine/server/storage/datastore/DsProjectionStorageDelegate.java
```diff
@@ -0,0 +1,142 @@
+/*
+ * Copyright 2017, TeamDev Ltd. All rights reserved.
```
armiol: It's 2018.

armiol (changes requested):
@dmdashenkov please see my comments.

dmdashenkov:
@armiol, PTAL again.

armiol (approved):
@dmdashenkov LGTM.
Please make sure this version is published to the maven repo.

Also, in future we'll need to address multiple string concatenations (i.e. "+" operations) that we have now. They don't look very good.

PR #69, opened 2018-03-12
Fix aggregate reads

dmdashenkov:
Before, the `historyBackward(...)` reads filtered out records marked as `archived` or `deleted`.
Now this behavior is suspended, i.e. any record is accessible by its ID in spite the lifecycle.


dmdashenkov:
@armiol, PTAL.

armiol (approved):
@dmdashenkov LGTM.

PR #84, opened 2018-06-08
Add support for Enum Entity Columns

dmitrykuzmin:
This PR is a part of https://github.com/SpineEventEngine/core-java/pull/698.

As the support for enumerated entity columns has been added to the `core-java`, we need to add support for enumerated column filters to all storage implementation.

In case of `gcloud-java`, the enumerated filter value will be converted to the `com.google.cloud.datastore.Value` of integer or string type for the further usage.

Also now adapter of column filters supports `null` filter values (converting them to datastore `NullValue`), although `null` filter values themselves are not yet supported in the `core-java`.

`gcloud-java` version advances to `0.10.41-SNAPSHOT`.


dmitrykuzmin:
@armiol PTAL.

datastore/src/test/java/io/spine/server/storage/datastore/DsFiltersShould.java
```diff
@@ -164,6 +168,28 @@ public void generate_filters_from_empty_params() {
         assertSize(0, filters);
     }
 
+    //TODO:2018-06-08:dmytro.kuzmin: re-write without mocks when null column filters are available.
```
armiol, line 171: This has to be a reference to a `core-java` issue, which tells to implement `null` column filters.

armiol (approved):
@dmitrykuzmin LGTM with a single comment to address before merging.

PR #85, opened 2018-06-29
Remove `web` from the Module

dmdashenkov:
In this PR we remove the codebase of the Spine Web module. Removed modules:
 - `web`;
 - `firebase-web`;
 - `client-js`;
 - `client-js-proto`;
 - `web-tests`.

Also, in this PR we migrate this repository to the latest Spine:
 - added `config` submodule and updated dependencies according to definitions in `config`;
 - removed `final` from local variables; 
 - migrated to Java 8 API where possible;
 - migrated to JUnit 5 API and code conventions.


dmdashenkov:
@armiol, PTAL.

datastore/src/main/java/io/spine/server/storage/datastore/Indexes.java
```diff
@@ -24,26 +24,30 @@
 import com.google.cloud.datastore.EntityQuery;
 import com.google.cloud.datastore.Key;
 import com.google.cloud.datastore.Query;
-import com.google.common.base.Function;
 import com.google.common.collect.Iterators;
+import com.google.common.collect.Streams;
+import io.spine.server.storage.Storage;
```
armiol: I think this import statement is only used for `{@linkplain}`s and therefore the FQNs in the links should be restored and the import should be removed.

armiol (approved):
@dmdashenkov LGTM with a single comment to address.
Also, it would be great to move the nested `Given`s to a separate package.

Issue #95, opened 2018-11-12
`DsAggregateStorage` overwrites any previously created snapshot with a new one


yuri-sergiichuk:
Now, `DsAggregateStorage` overwrites any previously stored snapshot of a particular entity.

The record ID for a snapshot is generated using a static `SNAPSHOT` string concatenated with a stringified version of an entity ID. It means, that when it is time to store a new snapshot the Datastore record ID for the newly created snapshot is simply equal to the previously stored snapshot.

The desired behavior is to store multiple entity snapshots without actually overriding the previously-created ones.

armiol:
Closing as resolved.

Issue #90, opened 2018-09-06
`DsAggregateStorage` overwrites event count after last snapshot for different aggregates with the same ID


dmytro-grankin:
Now, `DsAggregateStorage`, overwrites an event count after the last snapshot for different aggregates with the same IDs.

The event count is stored in the `google.protobuf.Int32Value` kind. The ID of the count consists of `EVENTS_AFTER_SNAPSHOT_` prefix and a stringified value of an aggregate ID.

So, when two aggregates (e.g. `TaskRoot` and `TaskPart`) have the same ID (`task-123`), they have the same ID for the event count after the last snapshot.

To solve the problem, an ID of an event count should additionally contain a state type name of an aggregate. So, the ID should look like `EVENTS_AFTER_SNAPSHOT_foo.bar.StateType_task-123`.

The same problem exists for the aggregate lifecycle flags. Also, an ID of an aggregate lifecycle flags includes `EVENTS_AFTER_SNAPSHOT_` prefix, which is strange.

dmdashenkov:
Fixed in #96.

PR #96, opened 2018-11-22
Fix event count and snapshot records being overwritten unexpectedly

dmitrykuzmin:
This PR addresses issues #90 and #95.

Previously, the IDs of event count records and Aggregate snapshot records could be unexpectedly overwritten because their IDs weren't specific enough.

This PR changes the ID format as follows:

1. For the event records, the Aggregate state type is appended to the ID, for example,
`EVENTS_AFTER_SNAPSHOT_spine.test.aggregate.Project_{"id":"randomId"}`.
2. For the snapshots, the `Timestamp` of the snapshot is appended to the ID, for example:
`SNAPSHOT_{"id":"randomId"}_2018-11-22T18:14:16.267Z`

Also, the lookup by the old-format IDs was implemented to ensure compatibility with the previously saved data.

This PR also migrates the `gcloud-java` tests to always executing on the local datastore emulator (even on CI). There is only one acceptance test left for the remote datastore to check if everything is set up correctly when building on CI.


dmitrykuzmin:
@dmdashenkov PTAL.

datastore/src/main/java/io/spine/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -224,6 +225,34 @@ protected RecordId toRecordId(I id) {
...
+     * Generates an identifier for the Datastore record which represents an Aggregate snapshot.
+     *
+     * @param id
+     *         an identifier of the {@code Aggregate}
+     * @return the identifier for the Datastore record
+     */
+    protected RecordId toSnapshotId(I id, Snapshot snapshot) {
+        String stringId = Stringifiers.toString(id);
+        String snapshotTimeStamp = Timestamps.toString(snapshot.getTimestamp());
+        String snapshotId = SNAPSHOT + '_' + stringId + '_' + snapshotTimeStamp;
```
dmdashenkov: Please `String.format(...)` this value instead.

datastore/src/main/java/io/spine/server/storage/datastore/DsAggregateStorage.java
```diff
@@ -224,6 +225,34 @@ protected RecordId toRecordId(I id) {
...
+     * Generates an identifier for the Datastore record which represents an event count after the
+     * last snapshot.
+     *
+     * @param id
+     *         an identifier of the {@code Aggregate}
+     * @return the identifier for the Datastore record
+     */
+    protected RecordId toEventCountId(I id) {
+        String stringId = Stringifiers.toString(id);
+        String datastoreId = EVENTS_AFTER_LAST_SNAPSHOT_PREFIX + stateTypeName + '_' + stringId;
```
dmdashenkov: Same as above.

datastore/src/test/java/io/spine/server/storage/datastore/DatastoreWrapperTest.java
```diff
@@ -180,6 +184,38 @@ void testBigBulkRead() throws InterruptedException {
         }
     }
 
+    @Nested
+    class Remote {
```
dmdashenkov, line 197: Please document on the purpose of this suite.

datastore/src/test/java/io/spine/server/storage/datastore/given/DsAggregateStorageTestEnv.java
```diff
@@ -0,0 +1,39 @@
...
+import io.spine.server.aggregate.Aggregate;
+import io.spine.test.aggregate.ProjectId;
+
+public final class DsAggregateStorageTestEnv {
+
+    private DsAggregateStorageTestEnv() {
+    }
+
+    public static class ProjectAsStringAggregate
+            extends Aggregate<ProjectId, StringValue, StringValueVBuilder> {
```
dmdashenkov: Can we not use `StringValue` as an aggregate state type?
It is a bit fragile, since we might forbid such usage in the future. In fact, there are already limitations for aggregates not marked as `(entity).kind = AGGREGATE`.

datastore/src/test/java/io/spine/server/storage/datastore/given/TestDatastores.java
```diff
@@ -31,84 +31,68 @@
...
         return Ci.INSTANCE.getService();
     }
-
     /**
      * Obtains ProjectId of the test environment.
      */
     public static ProjectId projectId() {
         return TEST_PROJECT_ID;
     }
-
```
dmdashenkov: Please restore empty lines in this file. It makes it a bit more readable when there is no IDE to collapse method bodies.

dmdashenkov (commented):
@dmitrykuzmin, please see my comments for now.
Also, mind the upcoming [changes](https://github.com/SpineEventEngine/core-java/pull/892) in the storage API.

dmitrykuzmin:
@dmdashenkov PTAL.

dmdashenkov (approved):
@dmitrykuzmin, LGTM.

PR #98, opened 2019-01-21
Fix publishing

dmdashenkov:
In this PR we fix the broken publishing scripts by pulling the new versions of the scripts from `config`.

The version yet to be published is `1.0.0-pre3`.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #100, opened 2019-01-31
Expose Datastore client lib as an API

dmdashenkov:
This library depends on the Google Datastore client lib.

The API of the library expects the user to be able to use the Datastore lib as well. Thus, we make the Datastore client lib part of the API of this library by exposing it as a transitive dependency.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #103, opened 2019-02-15
Fix duplicate results on query with `EITHER` filters

### LINKED ISSUE #649: Inconsistent query results using `CompositeColumnFilter` with `EITHER` operator
For example, we want to query entities with any lifecycle flags and use the following filter:

```java
CompositeColumnFilter anyFlagsFilter = ColumnFilters.either(eq("archived", false),
                                                            eq("archived", true),
                                                            eq("deleted", false),
                                                            eq("deleted", true));
```

And the query itself:

```java
final Query query = actorRequestFactory.query()
                                       .select(entityClass)
                                       .where(anyFlagsFilter)
                                       .build
```

Results for such a query differs between `DsRecordStorage` and `InMemoryRecordStorage`.

Suppose, there is only one entity of the class and it has `archived` lifecycle flag set to `true` and `deleted` set to `false`. 

Even though only one entity is stored, `DsRecordStorage` will return 2 entities. In this case, `DsRecordStorage` makes 4 Datastore queries using each subfilter from `CompositeColumnFilter`, then merges results of these queries and return the merged result.

So, the entity will be returned firstly from Datastore query, which is based on `eq("archived", true)` subfilter and secondly from the query, which is based on `eq("deleted", false)` subfilter.

`InMemoryRecordStorage` returns only one entity for this query, without a duplication as for `DsRecordStorage`.

Javadoc for `ColumnFilters.either(...)` says that `A record is considered matching this filter if it matches at least one of the composite Column filters.` This comment corresponds to the behaviour of `InMemoryRecordStorage` and is convenient and logical. So, `DsRecordStorage` should behave in the same way and should not return duplications.

dmitrykuzmin:
This PR fixes https://github.com/SpineEventEngine/core-java/issues/649.

Google Cloud Datastore does not support `OR` composite filters so we emulate them by running multiple queries and concatenating the results.

When the query is ambigous i.e. multiple filters from `either` condition could be matched, we would return the same `Entity` more than once (one instance per match).

Now, the results produced by the Datastore lookup will be `distinct`. There is not much fancier we can do considering multiple Datastore Query [limitations](https://cloud.google.com/datastore/docs/concepts/queries#restrictions_on_queries).

### DIFF
--- a/datastore/src/main/java/io/spine/server/storage/datastore/DsLookupByIds.java
+++ b/datastore/src/main/java/io/spine/server/storage/datastore/DsLookupByIds.java
@@ -48,8 +48,9 @@ import static java.util.stream.Collectors.toList;
 /**
  * An {@code Entity} lookup in Google Datastore using {@code Entity} identifiers.
  *
- * @implNote Lookup is performed by reading all the entities with Datastore Keys matching
- *         the provided IDs first and then applying other query constraints in-memory.
+ * @implNote
+ * Lookup is performed by reading all the entities with Datastore Keys matching
+ * the provided IDs first and then applying other query constraints in-memory.
  */
 final class DsLookupByIds<I> {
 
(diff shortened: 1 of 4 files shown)


dmitrykuzmin:
@armiol PTAL.

armiol (approved):

PR #114, opened 2019-06-20
Stackdriver-based Trace API implementation

dmdashenkov:
This PR introduces the implementation of the new [Spine Trace API](https://github.com/SpineEventEngine/core-java/pull/1090) based on [Stackdriver Trace](https://cloud.google.com/trace/docs/).

When a command/an event is handled, a new span is created. Spans which represent the propagation of a single signal tree are grouped into a single trace.
For example, if the command `StartGroupImport` cased one event `GroupImportStarted` and many events `MemberImported`, some of which in turn caused several `UserCreated`, all of those commands and events belong to the same signal tree (with the root in `StartGroupImport`) and thus are displayed on the same trace.

The trace below shows an aggregate part `TaskPart` handle a `CreateTask` command and a `TaskViewProjection` handle the resulting `TaskCreated` event.

Screen Shot 2019-06-27 at 15 07 40

Spans include such information as the IDs of the signal and the receiver entity, the name of the bounded context to which the entity belongs, and the associated tenant ID.

The screenshot below shows the detailed info about a single span.
Screen Shot 2019-06-27 at 15 07 58


dmdashenkov:
@armiol, PTAL.

stackdriver-trace/src/main/java/io/spine/server/trace/stackdriver/SignalSpan.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.stream.Stream;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.protobuf.AnyPacker.unpack;
+import static java.lang.String.format;
+
+/**
+ * A span based on a signal.
+ *
+ * <p>Signal messages may be processed once or many times by different entities. Each such
```
armiol: Can you please tell more on the subject. I also don't think that `each such` is a good thing.

armiol (commented):
@dmdashenkov LGTM with the comment.

armiol:
@dmdashenkov also, I am not sure it's a good idea to have the span names that long. In particular, why the project ID is always included? Let's discuss vocally.

This PR description lacks a screenshot of the Google Console. Otherwise, we could see if the long names are OK.

dmdashenkov:
@armiol, PTAL.

dmdashenkov:
The AppVeyor build [is not yet configured](#115), thus it fails.

stackdriver-trace/src/main/java/io/spine/server/trace/stackdriver/SignalSpan.java
```diff
@@ -42,20 +43,36 @@
...
 
     private static final int SPAN_DISPLAY_NAME_LENGTH = 128;
+
     private final BoundedContextName context;
     private final Signal<?, ?, ?> signal;
     private final MessageId receiver;
+    private final EntityTypeName receiverType;
 
-    SignalSpan(BoundedContextName context, Signal<?, ?, ?> signal, MessageId receiver) {
+    protected SignalSpan(BoundedContextName context,
```
armiol, line 76: Please document.

stackdriver-trace/src/main/java/io/spine/server/trace/stackdriver/TraceService.java
```diff
@@ -28,7 +28,15 @@
  * <p>Depending on the implementation, the client may send queries in the same thread or
  * asynchronously.
  */
-interface TraceService extends AutoCloseable {
+public interface TraceService extends AutoCloseable {
 
+    /**
+     * Performs the given {@code BatchWriteSpansRequest}.
```
armiol: I think it's `Executes`.

stackdriver-trace/src/main/java/io/spine/server/trace/stackdriver/SignalSpan.java
```diff
@@ -0,0 +1,205 @@
...
+        private EntityTypeName receiverType;
+
+        /**
+         * Prevents direct instantiation.
+         */
+        private Builder() {
+        }
+
+        public Builder setContext(BoundedContextName context) {
+            this.context = checkNotNull(context);
```
armiol, line 189: We don't seem to test these for `null` values.

armiol (commented):
@dmdashenkov please see my comments. LGTM otherwise.

dmdashenkov:
@armiol PTAL again.

armiol (approved):

PR #117, opened 2019-07-09
Remove `TenantConverterRegistry`

armiol:
The `DatastoreStorageFactory` configuration allows supplying a custom `NamespaceConverter`. It is used to put the `TenantId` values into accordance with the Datastore namespaces, in which the tenant-specific data resides.

Before this PR we had a singleton called `TenantConverterRegistry`, which could store several converters for each of the Datastore Project IDs used in the application. 

As of now, the `DatastoreStorageFactory` is, in fact, a property of `ServerEnvironment`. So there is the only instance of this factory per JVM. Therefore, it's impossible to configure it with multiple Datastore Project IDs. 

This PR removes the redundant singleton-registry of converters and propagates the one set by the framework user (or a framework default converter, if no custom set) down to the routines working with `Namespace`s.


armiol:
@dmdashenkov PTAL.

datastore/src/main/java/io/spine/server/storage/datastore/DatastoreStorageFactory.java
```diff
@@ -182,18 +184,12 @@ protected DsPropertyStorage createPropertyStorage(ContextSpec spec) {
     }
 
     private NamespaceSupplier createNamespaceSupplier(boolean multitenant) {
-        @Nullable String defaultNamespace;
         if (multitenant) {
             checkHasNoNamespace(datastore);
-            defaultNamespace = null;
+            return NamespaceSupplier.singleTenant();
```
dmdashenkov: This surely must be a typo: if `multitenant`, return single tenant supplier.

datastore/src/main/java/io/spine/server/storage/datastore/tenant/DefaultNamespaceConverter.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    protected String toString(TenantId tenantId) {
+        return forTenant(tenantId).toString(tenantId);
+    }
+
+    @Override
+    protected TenantId toTenantId(String namespace) {
+
```
dmdashenkov: Please remove this line.

dmdashenkov (changes requested):
@armiol, please see my comments.

datastore/src/main/java/io/spine/server/storage/datastore/DatastoreStorageFactory.java
```diff
@@ -186,9 +186,11 @@ protected DsPropertyStorage createPropertyStorage(ContextSpec spec) {
     private NamespaceSupplier createNamespaceSupplier(boolean multitenant) {
         if (multitenant) {
             checkHasNoNamespace(datastore);
-            return NamespaceSupplier.singleTenant();
-        } else {
             return NamespaceSupplier.multitenant(converterFactory);
+        } else {
+            String defaultNamespace = datastore.getOptions()
+                                               .getNamespace();
+            return NamespaceSupplier.singleTenant(defaultNamespace == null ? "" : defaultNamespace);
```
dmdashenkov: Please use `Strings.nullToEmpty`.

dmdashenkov (approved):

PR #120, opened 2019-07-25
Entity column API update

dmdashenkov:
With the new column API, there is no difference between the stored column name and the queries column name.

This resolves the issue with querying entities by IDs and columns (which did not work before, if the stored and queried column names did not match).


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #121, opened 2019-08-02
Return unique aggregate IDs from the `DsAggregateStorage.index()`

dmitrykuzmin:
This PR fixes the work of the `DsAggregateStorage.index()` method.

Previously, the method returned a lot of repetitive aggregate IDs, returning an ID for each `AggregateEventRecord` found in the storage. Now, the method will return only unique aggregate IDs.

This is done with the help of Datastore `ProjectionEntityQuery`.

As the `DatastoreWrapper` now supports structured queries for arbitrary objects (and not only `StructuredQuery<Entity>`), the `DsRecordStorage` index was also migrated - to Datastore `KeyQuery` to avoid consuming additional entity read operations.

The `core` and `base` dependencies are bumped to `1.0.0` to fetch the corresponding `AggregateStorage` [test](https://github.com/SpineEventEngine/core-java/pull/1133).

### DIFF
--- a/datastore/src/main/java/io/spine/server/storage/datastore/DatastoreWrapper.java
+++ b/datastore/src/main/java/io/spine/server/storage/datastore/DatastoreWrapper.java
@@ -267,11 +267,11 @@ public class DatastoreWrapper implements Logging {
      * @param query
      *         {@link Query} to execute upon the Datastore
      * @param <E>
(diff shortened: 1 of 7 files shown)


dmitrykuzmin:
@armiol PTAL.

armiol (approved):

PR #124, opened 2019-08-20
Custom namespaces for multitenant storages

dmdashenkov:
Previously, the Datastore storage factory limited usage of custom datastore namespaces to single-tenant storages, since the multitenancy mechanism is implemented via the Datastore namespaces.
However, this approach is not good enough anymore, since `InboxStorage` is always multitenant, effectively making it impossible to use custom namespaces at 

In this PR we provide a workaround for this issue. Now, multitenant storages may be created with a custom namespace. In this case, the provided namespace is used as a prefix for the tenant-based namespace.
The resulting Datastore namespace has the following format:
`<custom-namespace>.<tenant-ID-based-namespace>`.


dmdashenkov:
@armiol, PTAL.

datastore/src/main/java/io/spine/server/storage/datastore/tenant/PrefixedNsConverterFactory.java
```diff
@@ -0,0 +1,80 @@
...
+
+        private Converter(String prefix, NamespaceConverter delegate) {
+            super(prefix);
+            this.delegate = checkNotNull(delegate);
+        }
+
+        @Override
+        TenantId significantStringToTenantId(String namespace) {
+            return namespace.startsWith(SEPARATOR)
+                   ? delegate.convert(namespace.substring(1))
```
armiol: Let's use `SEPARATOR` somehow!

datastore/src/test/java/io/spine/server/storage/datastore/DatastoreStorageFactoryTest.java
```diff
@@ -109,6 +121,67 @@ void testClose() {
...
+        DsPropertyStorage storage = factory.createPropertyStorage(spec);
+        Key key = whiteForTenant(storage, tenant)
+                .setNamespace(namespace + ".V" + tenant.getValue())
+                .build();
+        Entity entity = datastore.get(key);
+        assertThat(entity).isNotNull();
+    }
+
+    @Test
+    @DisplayName("allow no namespaces for multitenant storages")
```
armiol: `no custom namespaces`

armiol (approved):
@dmdashenkov LGTM with two comments to address.

PR #125, opened 2019-08-28
Delivery improvements

armiol:
This changeset adds some precautions to using `DsShardedWorkRegistry` in a multi-node environment:

1. The shard index is now read using both the index value and the total number of shards. It prevents from confusing the indexes such as "1 of 10" and "1 of 42" with each other.
2. After picking up the session we now verify that this node has indeed written the contents to the registry.
3. The `pickUp()` call is now synchronized to avoid extra I/O of reading/writing the concurrent session records from the same instance.

### DIFF
--- a/datastore/src/main/java/io/spine/server/storage/datastore/DsShardedWorkRegistry.java
+++ b/datastore/src/main/java/io/spine/server/storage/datastore/DsShardedWorkRegistry.java
@@ -25,13 +25,15 @@ import com.google.cloud.datastore.LongValue;
 import com.google.cloud.datastore.Query;
 import com.google.cloud.datastore.StringValue;
 import com.google.cloud.datastore.TimestampValue;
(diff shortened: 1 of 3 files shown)


datastore/src/main/java/io/spine/server/storage/datastore/DsShardedWorkRegistry.java
```diff
@@ -80,9 +81,24 @@ public DsShardedWorkRegistry(DatastoreStorageFactory factory) {
         }
         ShardSessionRecord ssr = newRecord(index, nodeId);
         write(ssr);
+        boolean writeConsistent = checkConsistency(index, ssr);
```
dmdashenkov, line 84: Please get rid of the abbreviation.

dmdashenkov (approved):
@armiol, LGTM with a single comment.

PR #127, opened 2019-09-04
Release only non-empty sessions

dmdashenkov:
Before this PR, when releasing shard processing sessions, we cleaned up all the sessions regardless of whether or not a session is active (has a processing node) or not. Now, we only release the active sessions.

See the [same change](https://github.com/SpineEventEngine/core-java/pull/1157) in `core-java`.

### DIFF
--- a/version.gradle
+++ b/version.gradle
@@ -25,12 +25,12 @@
  * {@code .config/gradle/dependencies.gradle}.
  */
 
-def final SPINE_VERSION = '1.0.3'
+def final SPINE_VERSION = '1.0.7-SNAPSHOT'
 
 ext {
     versionToPublish = SPINE_VERSION
     
-    spineBaseVersion = SPINE_VERSION
+    spineBaseVersion = '1.0.3'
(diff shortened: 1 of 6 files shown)


dmdashenkov:
@alexander-yevsyukov, PTAL.

dmdashenkov:
@alexander-yevsyukov, PTAL again.

version.gradle
```diff
@@ -25,12 +25,12 @@
...
  */
 
-def final SPINE_VERSION = '1.0.3'
+def final SPINE_VERSION = '1.0.7-SNAPSHOT'
 
 ext {
     versionToPublish = SPINE_VERSION
     
-    spineBaseVersion = SPINE_VERSION
+    spineBaseVersion = '1.0.3'
```
alexander-yevsyukov: Please use the recent version of `base`.

alexander-yevsyukov (approved):
LGTM with one comment related to the version of `base`.

PR #132, opened 2019-10-08
Thread-safe transactions

dmdashenkov:
In this PR we introduce an API for thread-safe Datastore transactions.

Previously, transactions were handled by `DatastoreWrapper`. When required, the wrapper substituted the `Datastore` for a transaction and executed all the requests upon that transaction.
This approach caused concurrency issues.
Now that API is deprecated. It should be removed in subsequent releases.

The new API for transactions allows managing a single Datastore transaction independently of all the others.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #133, opened 2019-10-28
The `(column)` option

dmitrykuzmin:
This PR propagates the changes in https://github.com/SpineEventEngine/core-java/pull/1184 to the `gcloud` storage.

The main change is that now there is no more `ColumnTypeRegistry`. The column value conversion is instead done with the help of `ColumnMapping` - a new structure which is similar to the old type registry but conforms with the new column API.

The Spine version advances to `1.1.13`.


dmitrykuzmin:
@armiol, @dmdashenkov PTAL.

dmdashenkov (approved):
@dmitrykuzmin, LGTM.

PR #139, opened 2019-12-25
Improve DatastoreWrapper logging

yuri-sergiichuk:
In this PR I have cleaned up the DatastoreWrapper logs as the previous log had no useful information at all:
```
FINE: Using namespace `hhcolorlab-v2`.
Dec 25, 2019 4:26:33 PM io.spine.server.storage.datastore.DatastoreWrapper currentNamespace
FINE: Using namespace `hhcolorlab-v2`.
Dec 25, 2019 4:26:33 PM io.spine.server.storage.datastore.DatastoreWrapper currentNamespace
FINE: Using namespace `hhcolorlab-v2`.
```

The new logs give more content and use a lower logging level as well.

Also, as a part of the PR, I have bumped our Datastore and Trace API versions.

### DIFF
--- a/datastore/src/main/java/io/spine/server/storage/datastore/DatastoreWrapper.java
+++ b/datastore/src/main/java/io/spine/server/storage/datastore/DatastoreWrapper.java
@@ -574,7 +574,7 @@ public class DatastoreWrapper implements Logging {
      */
     private Iterator<Entity> readBulk(List<Key> keys) {
         int pageCount = keys.size() / MAX_KEYS_PER_READ_REQUEST + 1;
-        _debug().log("Reading a big bulk of entities synchronously. The data is read as %d pages.",
+        _trace().log("Reading a big bulk of entities synchronously. The data is read as %d pages.",
                      pageCount);
         int lowerBound = 0;
         int higherBound = MAX_KEYS_PER_READ_REQUEST;
--- a/datastore/src/main/java/io/spine/server/storage/datastore/DatastoreWrapper.java
+++ b/datastore/src/main/java/io/spine/server/storage/datastore/DatastoreWrapper.java
@@ -278,7 +278,7 @@ public class DatastoreWrapper implements Logging {
                 query.toBuilder()
                      .setNamespace(namespace.getValue())
                      .build();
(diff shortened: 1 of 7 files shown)


datastore/src/main/java/io/spine/server/storage/datastore/DatastoreWrapper.java
```diff
@@ -274,11 +273,13 @@ public void createOrUpdate(Collection<Entity> entities) {
      * @see DatastoreReader#run(Query)
      */
     public <R> DsQueryIterator<R> read(StructuredQuery<R> query) {
-        Namespace namespace = currentNamespace();
+        Namespace namespace = namespaceSupplier.get();
         StructuredQuery<R> queryWithNamespace =
                 query.toBuilder()
                      .setNamespace(namespace.getValue())
                      .build();
+        _trace().log("Reading entities of `%s` kind with `%s` filter in `%s` namespace.",
```
dmdashenkov:
```suggestion
        _trace().log("Reading entities of kind: `%s` in namespace: `%s` .",
```

dmdashenkov: Please also adjust the arguments line.

dmdashenkov: Please also apply to the other logs in this file.

dmdashenkov (changes requested):

dmdashenkov:
@yuri-sergiichuk, let's discuss what values we can and cannot log. For instance, I'm not sure we should log filter values.

datastore/src/main/java/io/spine/server/storage/datastore/DatastoreWrapper.java
yuri-sergiichuk: As discussed vocally, we'd better avoid additional special characters in logs as it generally makes logs filtering harder in e.g. Stackdriver

dmdashenkov (approved):
@yuri-sergiichuk, LGTM.

PR #143, opened 2020-01-27
Expose transaction API

dmdashenkov:
This PR updates the API of the Cloud Datastore-based storage. In particular, we make it easier to work with transactions. The new methods of `TransactionWrapper` cover all the necessary database operations.

Also, in this PR, we update the format of `InboxMessage` storage. Previously, `InboxMessage`s had a Timestamp property `when_received`. However, with the latest time-related changes in the framework, the precision of the Datastore Timestamp is now not enough for us. Thus, we introduce a new column `received_at`, which contains the number of nanoseconds, elapsed since the Unix epoch start. Though nanoseconds are, strictly speaking, a lie, this is good enough for our technical purposes.

As another side effect of the time-related changes, the standard idiom for querying entities by a time-based property has changed. Instead of passing an `equals` filter, it is recommended to
 1) pass an `equals` filter with a value rounded down to a whole microsecond;
 2) or pass a greater than - less than filter with a span of ± 1 second.

### DIFF
--- a/datastore/src/main/java/io/spine/server/storage/datastore/DsInboxStorage.java
+++ b/datastore/src/main/java/io/spine/server/storage/datastore/DsInboxStorage.java
@@ -41,6 +41,7 @@ import static com.google.cloud.datastore.StructuredQuery.CompositeFilter.and;
 import static com.google.cloud.datastore.StructuredQuery.OrderBy.asc;
 import static com.google.cloud.datastore.StructuredQuery.PropertyFilter.eq;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.protobuf.util.Timestamps.toNanos;
 import static io.spine.server.delivery.InboxMessageStatus.TO_DELIVER;
 
 /**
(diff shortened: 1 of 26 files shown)


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@armiol, PTAL.

datastore/src/main/java/io/spine/server/storage/datastore/DsInboxStorage.java
```diff
@@ -147,8 +148,10 @@ InboxMessageId idOf(InboxMessage message) {
                                    .toString());
         }),
 
-        whenReceived("when_received", (m) -> {
-            return TimestampValue.of(fromProto(m.getWhenReceived()));
+        receivedAt("received_at", (m) -> {
+            Timestamp timestamp = m.getWhenReceived();
+            long epochNanos = timestamp.getSeconds() * NANOS_PER_SECOND + timestamp.getNanos();
```
armiol: Please use `Timestamps.toNanos(..)`.

datastore/src/main/java/io/spine/server/storage/datastore/TransactionWrapper.java
```diff
@@ -23,30 +23,87 @@
...
+import java.util.List;
 import java.util.Optional;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static java.util.Optional.ofNullable;
 
 /**
  * A Cloud Datastore transaction wrapper.
+ *
+ * @implNote The wrapper provides API for basic operations which can be done transactionally.
```
armiol: Please rephrase (as discussed) to emphasize on the Datastore limits.

armiol (approved):
@dmdashenkov LGTM with two minor comments to address.

PR #146, opened 2020-02-27
Implement catch-up storage and use transactions for `Delivery` routines

armiol:
This changeset brings the support of catch-up operations to the applications running on top of Google Datastore.

**`DsCatchUpStorage`**

A new`DsCatchUpStorage` is now responsible for storing the state of the catch-up processes. 

Please note, in its work, this implementation is not using Datastore transactions. As the framework has no goal to handle the concurrent access to the state of a particular catch-up process, the code has been left fairly simple.

**Modes for `DsInboxStorage`**

`DsInboxStorage` has been modified to support transactional operations when operating on top of Firestore in Datastore mode. 

In particular, in this mode, the `InboxMessage` records are stored as children of `ShardIndex`. That allows querying and modifying them in the scope of a Datastore transaction. For a framework user, it means that all operations performed during the message delivery and catch-up are executed strictly in the desired order, as Datastore provides serializable isolation for its transactions.

Legacy native Datastore operational mode is also supported, but due to its limits and risk of contention on record operations, no transactional work is performed. 

By default, the `DatastoreStorageFactory` is instantiated with the `DsInboxStorage` in **native** mode.  It allows running the unit tests on the Datastore emulator, which so far only acts like a Datastore in native mode.

To switch to the Firestore in Datastore mode, an SPI user should override the 
`DatastoreStorageFactory.inboxStorageWith(multitenant, datastore)` method and return the `DsInboxStorage` created with `DatastoreMode.FIRESTORE_AS_DATASTORE` value:

```java
    @Override
    protected InboxStorage inboxStorageWith(DatastoreWrapper wrapper,
                                            boolean multitenant) {
        return new DsInboxStorage(wrapper, 
                                  multitenant,
                                  DatastoreMode.FIRESTORE_AS_DATASTORE);
    }
```

Please note that other storage implementations so far do not require the specific support of Datastore operational mode. Thus, their code has remained intact.

**Improved shard pick-up**

The `DsShardedWorkRegistry` has been updated to perform the shard pick-up in the scope of a Datastore transaction. It allows dealing with the potential concurrency issues in a more natural platform-specific way, rather than trying to read-write-read the same record as previously.

This feature is intended to work notwithstanding the Datastore mode.

**Other changes**

The API of `TransactionalWrapper` and `DsMessageStorage` was updated to allow handling the parent-children (ancestor) relations for the stored records.

The sample Datastore index files were updated according to the latest changes. There is also a new sample index set for Firestore in Datastore mode. For more details, see `/datastore/config/README.md`.

The library version is set to `1.4.11`.


datastore/config/README.md
```diff
@@ -1,7 +1,10 @@
...
- - Contain the Spine internal types config.
- - Contain your custom `Aggregate` types config.
- 
-For more details, please see `index.yaml`.
+Depending on the mode in which your Datastore runs one should choose either
+
+- `index-datastore-native.yaml` for the Datastore in native mode;
+- `index-firestore-in-datastore-mode.yaml` for the Firestore in Datastore mode.
+
+Each of the configurations lists the indexes for the system records (e.g. `Event`) and the samples for the user-defined entities.
```
dmdashenkov: Please check the length of this line.

datastore/src/main/java/io/spine/server/storage/datastore/DsInboxStorage.java
```diff
@@ -181,4 +310,156 @@ public String columnName() {
             return getter;
         }
     }
+
+    /**
+     * A behavior of the storage depending on the Datastore mode.
+     */
+    private interface Behavior {
```
dmdashenkov: `Behavior` seems to be too broad. How about `Client`?

datastore/src/main/java/io/spine/server/storage/datastore/DsMessageStorage.java
```diff
@@ -44,7 +46,7 @@
 import static java.util.stream.Collectors.toList;
 
 /**
- * An abstract base for storages operating plain {@link Message}s in Datastore.
+ * An abstract base for storage operating plain {@link Message}s in Datastore.
```
dmdashenkov:
```suggestion
 * An abstract base for a storage persisting plain {@link Message}s in Datastore.
```

datastore/src/main/java/io/spine/server/storage/datastore/DsMessageStorage.java
```diff
@@ -175,43 +275,50 @@ public void writeAll(Iterable<M> messages) {
...
+    protected final void writeTransactionally(M message) {
+        checkNotNull(message);
+
+        try (TransactionWrapper tx = newTransaction()) {
+            Entity entity = toEntity(message);
+            tx.createOrUpdate(entity);
+            tx.commit();
+        } catch (RuntimeException e) {
+            throw newIllegalStateException(e,
+                                           "Error writing a `%s` in transaction.",
```
dmdashenkov:
```suggestion
                                           "Error writing a `%s` in a transaction.",
```

datastore/src/test/java/io/spine/server/storage/datastore/TransactionWrapperTest.java
```diff
@@ -139,6 +141,52 @@ void readWrite() {
...
+        creationTx.commit();
+
+        Entity firstEntity = entities.remove(0);
+        TransactionWrapper singleDeletionTx = datastore.newTransaction();
+        Key toDelete = firstEntity.getKey();
+        singleDeletionTx.delete(toDelete);
+        singleDeletionTx.commit();
+
+        TransactionWrapper firstReadTx = datastore.newTransaction();
+        Truth8.assertThat(
```
dmdashenkov: Please add a static import for that.

dmdashenkov: The beauty of static imports is in the fact that you *can* import methods with the same name, as long as they don't clash parameter-wise.

datastore/src/test/java/io/spine/server/storage/datastore/TransactionWrapperTest.java
```diff
@@ -189,23 +237,30 @@ void notRollbackIfNotActive() {
...
-                }
-            }));
-        service.awaitTermination(5, SECONDS);
+        List<Callable<Key>> tasks =
+                keys.stream()
+                    .map(key -> Entity
+                            .newBuilder(key)
+                            .set("a", newUuid())
+                            .build())
+                    .map(entity -> (Callable<Key>) () -> {
```
dmdashenkov: This is too deep into the rabbit hole. Please extract the inner lambda.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.1.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.2-bin.zip
```
dmdashenkov: Please use 6.2.1.

dmdashenkov (changes requested):

dmdashenkov (approved):

PR #150, opened 2020-06-24
Add pubsub module

yuri-sergiichuk:
In this PR I have added a new `pubsub` module with a reusable `PubsubPushRequest` wrapper around the Pub/Sub push requests.

The reasoning for such a proto is that we've already created and recreated that same functionality internally 3 or 4 times and it [doesn't look like](https://github.com/googleapis/java-pubsub/issues/23) the respective proto definition may appear in the official repo any time soon.

The defined proto could be used for endpoints that are used with Pub/Sub Push subscriptions in order to easily transform incoming JSON data into a respective Protobuf definition.


pubsub/src/main/proto/spine/messaging/pubsub/pubsub_push.proto
```diff
@@ -0,0 +1,66 @@
+/*
```
armiol: I am not sure why this file is called `pubsub_push.proto`. `pubsub` is present in both package name and library names. 

`push.proto` is enough.

armiol (approved):
@yuri-sergiichuk LGTM with a comment to address before merging.

PR #164, opened 2021-06-10
Update to 2.x

armiol:
This PR updates this library to the new Storage API introduced in Spine Server 2.0.

Here is a breakdown of the changes made.

### Support of new Query DSL

Support of new Query DSL introduced in Spine 2.0, including [deep nesting of predicates](https://github.com/SpineEventEngine/core-java/pull/1368).

### Configuration-over-implementation:

**Enabling transactions**

It is now possible to select some record types which will be written and read transactionally:

```java
DatastoreStorageFactory.newBuilder()
           // ...
           .enableTransactions(ProjectDetails.class)
```

**Ancestor-child layout**

For each type of stored records, it is now possible to configure an ancestor-child Entity Group. This may be handy to leverage additional consistency by running transactions for such Entity Groups. 

While Datastore supports [multiple level of parents](https://cloud.google.com/datastore/docs/concepts/entities#ancestor_paths), this changeset now introduces just a single level of nesting, which seems sufficient at this point.

Library users may achieve this by customizing a record layout:

```java
DatastoreStorageFactory.newBuilder()
           // ...
           // Store each `InboxMessage` in an Entity Group under a `ShardIndex` parent:
           .organizeRecords(InboxMessage.class, new InboxStorageLayout())  
```
**Custom storage implementation**

Sometimes, it is required to customize how records of some particular type are stored. Previously, library users had to make several steps, including the creation of their own `StorageFactory`. 

Now, it can be achieved as simple as this:

```java
 DatastoreStorageFactory factory = DatastoreStorageFactory.newBuilder()
           // ...
           // Customize how `ProjectDetails` entity state is stored.
           .useEntityStorage(ProjectDetails.class, config -> myProjectionStorageWith(config))
           // ...
           // When dealing with plain Message records of `InboxMessage` type, use another custom storage.
           .useRecordStorage(InboxMessage.class, config -> myRecordStorageWith(config))
```

**Defaults**

By default, transactions are disabled for all storages except for `DsShardedWorkRegistry`. 

Also, if not told otherwise, all records are stored in a flat manner (e.g., no ancestor-child nesting is applied).

**BREAKING CHANGES**

Combined, an ability to enable transactions and organize records into ancestor-child Entity Groups *replace* the `DatastoreMode` setting, which was previously available in the library. The idea here is to allow library users to decide whether they really need transactions and grouping in their particular instance of Datastore, rather than having a single switch.

## Repackaging

Previously, the package structure was mostly flat, making the root package to over-inflate and collapse soon.

Therefore, in 2.x a major repackaging has been made. However, as to the public API, most of the major types such as `DatastoreStorageFactory` remained in place.

Along with the repackaging, lots of obsolete types were removed.

## Other Changes

Gradle has been updated to 6.9. Build scripts were migrated to Kotlin DSL.

Client SDKs of Google Clould products were updated to their latests versions (sadly, with some conflicts among _their_ dependencies, which was overcome by forcing the most recent versions).

The library version has been set to `2.0.0-SNAPSHOT.1`.


build.gradle.kts
```diff
@@ -24,144 +24,167 @@
...
     dependencies {
-        classpath(deps.build.gradlePlugins.errorProne) {
-            exclude(group = "com.google.guava")
+        classpath("io.spine.tools:spine-mc-java:$spineBaseVersion")
+    }
+
+    configurations.all {
+        resolutionStrategy {
+            force(
+                "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion",
```
dmdashenkov: I suggest `Kotlin.stdLib` and `Kotlin.stdLibCommon` instead. We could extract `Kotlin` into a variable for brevity.

build.gradle.kts
```diff
@@ -24,144 +24,167 @@
...
+            )
         }
-        classpath(deps.build.gradlePlugins.protobuf)
-        classpath("io.spine.tools:spine-model-compiler:$spineBaseVersion")
     }
 }
 
+repositories {
+    repositories.applyStandard()
+}
```
dmdashenkov: There's no need to wrap this `repositories.applyStandard()` into `repossitories { }`. You could use
```
repositories {
    applyStandard()
}
```
or
```
repositories.applyStandard()
```

build.gradle.kts
```diff
@@ -24,144 +24,167 @@
...
-    jacoco
-    @Suppress("RemoveRedundantQualifierName") // Cannot use imports here.
-    id("com.google.protobuf").version(io.spine.gradle.internal.Deps.versions.protobufPlugin)
-    @Suppress("RemoveRedundantQualifierName") // Cannot use imports here.
-    id("net.ltgt.errorprone").version(io.spine.gradle.internal.Deps.versions.errorPronePlugin)
+    io.spine.internal.dependency.Protobuf.GradlePlugin.apply {
+        id(id) version version
+    }
+    io.spine.internal.dependency.ErrorProne.GradlePlugin.apply {
+        id(id) version version
```
dmdashenkov, line 71: Cool, didn't know this works that way.

build.gradle.kts
```diff
@@ -24,144 +24,167 @@
...
-                force(Deps.build.guava, Deps.test.guavaTestlib)
-            }
+        with(Scripts) {
+            from(javacArgs(project))
+            from(modelCompiler(project))
+            from(projectLicenseReport(project))
+            from(slowTests(project))
+            from(testOutput(project))
+            from(javadocOptions(project))
+
```
dmdashenkov: Please kill this empty line.

build.gradle.kts
```diff
@@ -24,144 +24,167 @@
...
+    }
 
     java {
         sourceCompatibility = JavaVersion.VERSION_1_8
         targetCompatibility = JavaVersion.VERSION_1_8
     }
 
-    DependencyResolution.defaultRepositories(repositories)
-
+    repositories.applyStandard()
```
dmdashenkov: Given how we do this both for the root project and for the subprojects, let's just move it to `allprojects { }` and only call it once.

datastore/src/main/java/io/spine/server/storage/datastore/config/CreateEntityStorage.java
```diff
@@ -0,0 +1,39 @@
...
+
+import io.spine.server.entity.EntityRecord;
+
+/**
+ * A callback to create a storage for {@link io.spine.server.entity.Entity Entity} states.
+ *
+ * @param <I>
+ *         the type of {@code Entity} identifiers
+ */
+public interface CreateEntityStorage<I> extends CreateStorage<I, EntityRecord> {
```
dmdashenkov, line 38: Please add  `@Functionalnterface`.

datastore/src/main/java/io/spine/server/storage/datastore/config/CreateEntityStorage.java
```diff
@@ -0,0 +1,39 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.server.storage.datastore.config;
+
+import io.spine.server.entity.EntityRecord;
+
+/**
+ * A callback to create a storage for {@link io.spine.server.entity.Entity Entity} states.
```
dmdashenkov: Maybe, `A function ...`?

To me a callback is something about asynchronous API, etc., while this is a general purpose function.

datastore/src/main/java/io/spine/server/storage/datastore/query/DsFilters.java
```diff
@@ -0,0 +1,185 @@
...
+     * {@code (p1 & p2 & p4) | (p1 & p2 & p5) | (p1 & p3 & p4) | (p1 & p3 & p5)}.
+     *
+     * <p>The separate conjunctive groups (e.g. {@code (p1 & p2 & p4)}) in the result
+     * {@code Collection} are placed into a single {@link Filter} instances one per group.
+     *
+     * <p>In other words, the predicate expression is brought into the
+     * <a href="https://en.wikipedia.org/wiki/Disjunctive_normal_form">disjunctive normal form</a>.
+     *
+     * <p>Note that by the convention, the distinct
+     * {@link QueryPredicate} instances are considered to be joined by the {@code &}
```
dmdashenkov: It seems, `{@link QueryPredicate}` fits on the previous line.

datastore/src/main/java/io/spine/server/storage/datastore/query/DsFilters.java
```diff
@@ -0,0 +1,185 @@
...
+        List<StructuredQuery.Filter> filters =
+                parameters.stream()
+                          .map(param -> createFilter(param, mapping))
+                          .collect(toList());
+
+        checkState(!filters.isEmpty());
+        StructuredQuery.Filter first = filters.get(0);
+        StructuredQuery.Filter[] other = filters.subList(1, filters.size())
+                                                .toArray(new StructuredQuery.Filter[filters.size() - 1]);
+        StructuredQuery.Filter group = and(first, other);
```
dmdashenkov, line 169: Please create an overload of `and` which accepts an `Iterable` or a `List`, etc. This looks a bit too manual.

datastore/src/main/java/io/spine/server/storage/datastore/record/Entities.java
```diff
@@ -43,15 +42,18 @@
...
 import static com.google.common.base.Preconditions.checkState;
 import static io.spine.protobuf.AnyPacker.unpack;
-import static io.spine.server.storage.StateField.bytes;
 
 /**
  * Utility class for converting {@linkplain Message proto messages} into
  * {@linkplain Entity entities} and vise versa.
  */
-final class Entities {
+public final class Entities {
```
dmdashenkov, line 50: Can we mark this with `@Internal`?

datastore/src/main/java/io/spine/server/storage/datastore/record/RecordId.java
```diff
@@ -23,17 +23,17 @@
...
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static java.lang.String.format;
 
 /**
- * A wrapper type for the {@code String}-based record identifiers in the GAE Datastore.
+ * A wrapper type for the {@code String}-based record identifiers in Datastore.
  */
-final class RecordId extends DsIdentifier {
+public final class RecordId extends DsIdentifier {
```
dmdashenkov, line 36: Could this be `@Internal` too?

datastore/src/main/java/io/spine/server/storage/datastore/delivery/DsSessionStorage.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    private static StorageConfiguration<ShardIndex, ShardSessionRecord>
+    configureWith(DatastoreStorageFactory factory) {
+        DatastoreWrapper wrapper = factory.systemWrapperFor(DsSessionStorage.class, multitenant);
+        StorageConfiguration<ShardIndex, ShardSessionRecord> config =
+                StorageConfiguration.<ShardIndex, ShardSessionRecord>newBuilder()
+                        .withDatastore(wrapper)
+                        .withRecordSpec(newRecordSpec())
+                        .withContext(ContextSpec.multitenant("System"))
```
dmdashenkov: Do we define this somewhere in `core-java`? Seems strange to hardcode `"System"` (as well as the multitenancy setting).

datastore/src/main/java/io/spine/server/storage/datastore/delivery/RecordUpdate.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+interface RecordUpdate {
+
+    /**
+     * Decides in which update the existing record should result.
+     *
+     * @param previous
+     *         the previous record currently residing in the storage, or {@code null}
+     *         if there is no such record
+     * @return a version of the record to write to the storage,
```
dmdashenkov: Are the `@param` and `@return` mixed up?

datastore/src/main/java/io/spine/server/storage/datastore/delivery/InboxStorageLayout.java
```diff
@@ -0,0 +1,122 @@
...
+            ColumnName columnName = parameter.column()
+                                             .name();
+            if (columnName.equals(inbox_shard.name())) {
+                ShardIndex shard = (ShardIndex) parameter.value();
+                return Optional.of(shard);
+            }
+        }
+        for (QueryPredicate<InboxMessage> child : predicate.children()) {
+            Optional<ShardIndex> maybeResult = findShardIn(child);
+            if(maybeResult.isPresent()) {
```
dmdashenkov: Please add the space.

datastore/src/main/java/io/spine/server/storage/datastore/delivery/SessionRecordColumn.java
```diff
@@ -0,0 +1,73 @@
...
+     * Prevents this type from instantiation.
+     *
+     * <p>This class exists exclusively as a container of the column definitions. Thus it isn't
+     * expected to be instantiated. See the {@link RecordColumns} docs for more details on
+     * this approach.
+     */
+    private SessionRecordColumn() {
+    }
+
+    public static ImmutableList<RecordColumn<ShardSessionRecord, ?>> definitions() {
```
dmdashenkov, line 78: Seems like this could be package-private.

testutil-gcloud/src/main/java/io/spine/testing/server/storage/datastore/TestDatastoreStorageFactory.java
```diff
@@ -89,6 +97,7 @@ protected DatastoreWrapper createDatastoreWrapper(boolean multitenant) {
         return ImmutableSet.copyOf(allCreatedWrappers);
     }
 
+
```
dmdashenkov: Please undo this line.

datastore/src/main/java/io/spine/server/storage/datastore/DatastoreMedium.java
```diff
@@ -0,0 +1,196 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A common interface for those who wrap {@link DatastoreReaderWriter}s and provide their own
+ * API facade on top of them.
+ */
+public abstract class DatastoreMedium {
+
+    private final DatastoreReaderWriter datastoreReaderWriter;
```
alexander-yevsyukov: Can it be just `writer`, like “Lilu Dallas”?

datastore/src/main/java/io/spine/server/storage/datastore/config/CustomStorages.java
```diff
@@ -0,0 +1,124 @@
...
+
+    public static final class Builder
+            extends Settings.Builder<CreateStorage<?, ?>, CustomStorages, CustomStorages.Builder> {
+
+        private Builder() {
+            super();
+        }
+
+        @Override
+        Builder asThis() {
```
alexander-yevsyukov: Please consider renaming it to `self`.

datastore/src/main/java/io/spine/server/storage/datastore/config/RecordLayouts.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.Optional;
+
+import static io.spine.util.Exceptions.newIllegalArgumentException;
+
+/**
+ * The record layouts of storage implementations available
+ * to the {@linkplain io.spine.server.storage.datastore.DatastoreStorageFactory Datastore
+ * storage factory}.
+ *
+ * <p>This type is internal. The library users are able to supply the custom layout
```
alexander-yevsyukov: Shouldn't it be “a custom layout” instead of “the”?

datastore/src/main/java/io/spine/server/storage/datastore/config/RecordLayouts.java
```diff
@@ -0,0 +1,150 @@
...
+     *         the type of record identifiers
+     * @param <R>
+     *         the type of records, as a bounding generic parameter
+     * @return the record layout for the storage
+     * @throws IllegalArgumentException
+     *         if the record layout of wrong type was supplied on the configuration stage
+     * @implNote This method relies on the fact that the layout registered for the
+     *         storage of {@code <R>}-typed records with {@code <I>}-typed identifiers has the
+     *         corresponding generic parameters. Therefore, a plain class-cast is performed
+     *         when obtaining. If that fails, an exception is thrown.
```
alexander-yevsyukov: It's probably “the exception” since only `IllegalArgumentException` is announced in the documentation.

datastore/src/main/java/io/spine/server/storage/datastore/config/RecordLayouts.java
```diff
@@ -0,0 +1,150 @@
...
+    @SuppressWarnings({"unchecked", "OverlyBroadCatchBlock"})
+    private static <I, R extends Message>
+    RecordLayout<I, R> cast(RecordLayout<?, ?> raw, Class<?> domainType) {
+        RecordLayout<I, R> layout;
+        try {
+            layout = (RecordLayout<I, R>) raw;
+        } catch (Exception e) {
+            throw newIllegalArgumentException(e,
+                                              "Error using the provided storage layout" +
+                                                      " for type `%s`.", domainType);
```
alexander-yevsyukov: Please think of a more condensed layout. It's way too indented.

datastore/src/main/java/io/spine/server/storage/datastore/config/Settings.java
```diff
@@ -0,0 +1,143 @@
...
+ * An abstract base for types which declare various settings for storage implementations.
+ *
+ * @param <S>
+ *         the type of settings
+ * @param <T>
+ *         the type of concrete settings class
+ * @param <B>
+ *         the type of the builder of the {@code T}-typed settings class
+ */
+class Settings<S, T, B> {
```
alexander-yevsyukov: Assuming the Javadoc and the `abstract` `Builder` class below, isn't this class `abstract` too?

datastore/src/main/java/io/spine/server/storage/datastore/config/StorageConfiguration.java
```diff
@@ -0,0 +1,238 @@
...
+        public Builder<I, R> withTxSetting(TxSetting txSetting) {
+            this.txSetting = checkNotNull(txSetting);
+            return this;
+        }
+
+        public StorageConfiguration<I, R> build() {
+            checkNotNull(context, "`ContextSpec` must be set.");
+            checkNotNull(datastore, "`DatastoreWrapper` must be set.");
+            checkNotNull(recordSpec, "`DsRecordSpec` must be set.");
+            checkNotNull(columnMapping, "`ColumnMapping` must be set.");
```
alexander-yevsyukov: This block of checks suggests for having a helper function which takes a simple class name of a passed field.

alexander-yevsyukov (changes requested):
Please see my comments (122/213).

datastore/src/main/java/io/spine/server/storage/datastore/delivery/DsSessionStorage.java
```diff
@@ -0,0 +1,194 @@
...
+    private static DsEntitySpec<ShardIndex, ShardSessionRecord> newRecordSpec() {
+        MessageRecordSpec<ShardIndex, ShardSessionRecord> spec = messageSpec();
+        DsEntitySpec<ShardIndex, ShardSessionRecord> result = new DsEntitySpec<>(spec);
+        return result;
+    }
+
+    private static MessageRecordSpec<ShardIndex, ShardSessionRecord> messageSpec() {
+        @SuppressWarnings("ConstantConditions")     /* Protobuf getters never return `nulls`. */
+        MessageRecordSpec<ShardIndex, ShardSessionRecord> spec =
+        new MessageRecordSpec<>(ShardIndex.class,
```
alexander-yevsyukov: Please indent this block.

datastore/src/main/java/io/spine/server/storage/datastore/delivery/DsSessionStorage.java
```diff
@@ -0,0 +1,194 @@
...
+    @SuppressWarnings("OverlyBroadCatchBlock")  /* Treating all exceptions similarly. */
+    public final void write(ShardSessionRecord message) {
+        try (TransactionWrapper tx = newTransaction()) {
+            RecordWithColumns<ShardIndex, ShardSessionRecord> record = appendColumns(message);
+            Entity entity = entityRecordToEntity(record);
+            tx.createOrUpdate(entity);
+            tx.commit();
+        } catch (RuntimeException e) {
+            throw newIllegalStateException(e, "Error writing a `ShardSessionRecord`" +
+                    " in a transaction.");
```
alexander-yevsyukov: Can we have the whole message coming on the new line, please?

datastore/src/main/java/io/spine/server/storage/datastore/query/DsFilters.java
armiol: Yes, I think so too. But that's a type from the Datastore driver library.

datastore/src/main/java/io/spine/server/storage/datastore/record/Entities.java
armiol: This type seems to be a nice utility for those who needs to sanitize their Datastore storage manually. As long as we do some magic with our `bytes` field, this utility is the only appropriate way to read and write Datastore Entities outside of Spine's runtime.

datastore/src/main/java/io/spine/server/storage/datastore/record/RecordId.java
armiol: It's a part of our public API. See `FlatLayout` and other layouts for an example.

datastore/src/main/java/io/spine/server/storage/datastore/delivery/SessionRecordColumn.java
armiol: I have modified the documentation for this type to explain why it should remain `public`.

build.gradle.kts
armiol: I didn't either :) Just picked that up from one of our existing projects.

build.gradle.kts
```diff
@@ -24,144 +24,161 @@
...
+    @Suppress("RemoveRedundantQualifierName") // Cannot use imported things here.
+    io.spine.internal.dependency.ErrorProne.GradlePlugin.apply {
+        id(id) version version
+    }
+}
+
+spinePublishing {
+    targetRepositories.addAll(setOf(
+        PublishingRepos.cloudRepo
+        //, PublishingRepos.gitHub("core-java")
```
alexander-yevsyukov: Can we start publishing it to GitHub too?

datastore/src/main/java/io/spine/server/storage/datastore/delivery/DsSessionStorage.java
```diff
@@ -0,0 +1,209 @@
...
+    @SuppressWarnings("OverlyBroadCatchBlock")  /* Treating all exceptions similarly. */
+    public final void write(ShardSessionRecord message) {
+        try (TransactionWrapper tx = newTransaction()) {
+            RecordWithColumns<ShardIndex, ShardSessionRecord> record = appendColumns(message);
+            Entity entity = entityRecordToEntity(record);
+            tx.createOrUpdate(entity);
+            tx.commit();
+        } catch (RuntimeException e) {
+            throw newIllegalStateException(e,
+                                           "Cannot write a `ShardSessionRecord` in a transaction.");
```
alexander-yevsyukov: Please have both parameters on the second line. It would reduce the nesting.

datastore/src/main/java/io/spine/server/storage/datastore/delivery/SessionRecordColumn.java
```diff
@@ -0,0 +1,81 @@
...
+import io.spine.server.delivery.ShardSessionRecord;
+
+import static io.spine.query.RecordColumn.create;
+
+/**
+ * The definitions of record columns to store along with the {@link ShardSessionRecord}.
+ *
+ * @apiNote This type is made {@code public} to allow library users query the stored
+ *         {@code ShardSessionRecord}s via storage API, in case they need to read
+ *         the storage contents manually.
```
alexander-yevsyukov, line 42: Please also document on the reason for violating the Java standard for naming fields.

datastore/src/main/java/io/spine/server/storage/datastore/query/DsFilters.java
```diff
@@ -0,0 +1,184 @@
...
+    private static <R extends Message> Collection<StructuredQuery.Filter>
+    toDsFilters(QueryPredicate<R> predicate, FilterAdapter adapter) {
+        ImmutableList.Builder<StructuredQuery.Filter> result = ImmutableList.builder();
+        QueryPredicate<R> dnf = predicate.toDnf();
+        ColumnMapping<Value<?>> mapping = adapter.columnMapping();
+        if (dnf.operator() == AND) {
+            StructuredQuery.Filter group = handleConjunctiveGroup(dnf, mapping);
+            result.add(group);
+        } else {
+            checkState(dnf.allParams()
```
alexander-yevsyukov: Please extract this check into a method. It would 1) help to eliminate the funny wrapping, and 2) please consider documenting the reason for the check.

datastore/src/main/java/io/spine/server/storage/datastore/query/DsFilters.java
```diff
@@ -0,0 +1,184 @@
...
+                   "Children collection must be empty for a conjunctive predicate group.");
+        ImmutableList<SubjectParameter<?, ?, ?>> parameters = predicate.allParams();
+        List<StructuredQuery.Filter> filters =
+                parameters.stream()
+                          .map(param -> createFilter(param, mapping))
+                          .collect(toList());
+
+        checkState(!filters.isEmpty());
+        StructuredQuery.Filter first = filters.get(0);
+        StructuredQuery.Filter[] other = filters.subList(1, filters.size())
```
alexander-yevsyukov: Please extract `filters.subList()` thing into a method.

datastore/src/main/java/io/spine/server/storage/datastore/query/DsFilters.java
```diff
@@ -0,0 +1,184 @@
...
+    createFilter(SubjectParameter<?, ?, ?> parameter, ColumnMapping<Value<?>> mapping) {
+        checkNotNull(parameter);
+        checkNotNull(mapping);
+        Object paramValue = parameter.value();
+        ColumnTypeMapping<?, ? extends Value<?>> typeMapping = mapping.of(paramValue.getClass());
+        Value<?> value = typeMapping.applyTo(paramValue);
+        String columnName = parameter.column()
+                                     .name()
+                                     .value();
+        switch (parameter.operator()) {
```
alexander-yevsyukov: Can we have a conversion method, please?

datastore/src/main/java/io/spine/server/storage/datastore/query/DsLookupByIds.java
```diff
@@ -0,0 +1,136 @@
...
+        Stream<R> recordStream = stream.map(toMaskedRecord(mask()));
+        if (limit() != null && limit() > 0) {
+            recordStream = recordStream.limit(limit());
+        }
+        ImmutableList<R> result = recordStream.collect(toImmutableList());
+        return result;
+    }
+
+    private Predicate<Entity> columnPredicate() {
+        if (!(predicate().allParams()
```
alexander-yevsyukov: Please extract a variable for this “stairway” check.

datastore/src/main/java/io/spine/server/storage/datastore/query/DsLookupByQueries.java
```diff
@@ -0,0 +1,205 @@
...
+        Kind kind = spec().kind();
+        if (rootPredicate.isEmpty()) {
+            StructuredQuery<Entity> result = new QueryWithFilter(query, kind).withNoFilter();
+            return ImmutableList.of(result);
+        }
+
+        List<StructuredQuery<Entity>> queries = toDatastoreFilters(rootPredicate)
+                .stream()
+                .map(new QueryWithFilter(query, kind))
+                .collect(toList());
```
alexander-yevsyukov: Is there a reason for returning `List` rather than `ImmutableList` in this particular case?

alexander-yevsyukov (changes requested):
Please see my comments (136/212)

datastore/src/main/java/io/spine/server/storage/datastore/delivery/SessionRecordColumn.java
armiol: This is documented by `@RecordColumns` annotation by which this type is marked. We discussed that previously, and introduced it specifically for this reason.

However, as you have comments on this, we'll probably need to discuss this one more time — which is fine with me. I will think this over.

datastore/src/main/java/io/spine/server/storage/datastore/delivery/DsSessionStorage.java
armiol: I honestly tried to do so after your first comment on this matter :) But that's not what IDEA allows us to do with its auto-formatting.

datastore/src/test/java/io/spine/server/storage/datastore/given/DatastoreWrapperTestEnv.java
```diff
@@ -74,10 +59,21 @@ public static void ensureNamespace(String namespaceValue, Datastore datastore) {
...
+     */
     public static Datastore localDatastore() {
         return TestDatastores.local();
     }
 
+    /**
+     * Returns the {@code Datastore} instance connected to the Datastore instance
+     * running remotely.
+     *
+     * <p>The connection to a remote Datastore instance is performed via default service
```
dmdashenkov: As far as I remember "default" service account is the one used on AppEngine (and only there). I suggest just "performed via the service account".

datastore/src/test/java/io/spine/server/storage/datastore/given/TestEnvironment.java
```diff
@@ -59,10 +59,18 @@ public static boolean runsOnCi() {
         return onCi;
     }
 
+    /**
+     * Returns a new specification of a multi-tenant {@code BoundedContext} named after
```
dmdashenkov, line 63: I think, "multitenant" is a single word.

datastore/src/test/java/io/spine/server/storage/datastore/given/TestNamespaceSuppliers.java
```diff
@@ -24,17 +24,29 @@
...
 
+    /**
+     * Creates a single-tenant {@code NamespaceSupplier}.
+     */
     public static NamespaceSupplier singleTenant() {
         return NamespaceSupplier.singleTenant();
     }
 
+    /**
+     * Creates a multi-tenant {@code NamespaceSupplier}.
```
dmdashenkov, line 48: Same here.

build.gradle.kts
```diff
@@ -50,25 +49,21 @@ buildscript {
     configurations.all {
         resolutionStrategy {
             force(
-                "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion",
-                "org.jetbrains.kotlin:kotlin-stdlib-common:$kotlinVersion"
+                io.spine.internal.dependency.Kotlin.stdLib,
+                io.spine.internal.dependency.Kotlin.stdLibCommon
```
dmdashenkov: Please extract `Kotlin` into a `val`.

datastore/src/test/java/io/spine/server/storage/datastore/given/TestEnvironment.java
armiol: I think, not necessarily.

Please see [this](https://en.wikipedia.org/wiki/Talk%3AMultitenancy), [this](https://www.cisco.com/c/en/us/td/docs/switches/datacenter/dfa/configuration/b-dfa-configuration/multi-tenancy.pdf), and many others.

The idea I have here is that "multitenancy" is a long word, and in scope of our documentation it's much easier for eyes to catch this hyphen in "multi-tenancy", rather than parse this "..ltit.." in the middle of the long version.

datastore/src/main/java/io/spine/server/storage/datastore/query/DsFilters.java
armiol: You have helped me spotting a nasty bug!

Fixed now.

Thank you!

armiol:
@alexander-yevsyukov @dmdashenkov I have either addressed (or replied to) your comments. Please feel free to take another look.

datastore/src/test/java/io/spine/server/storage/datastore/query/DsFiltersTest.java
```diff
@@ -91,6 +96,41 @@ void testCompositeParams() {
...
+                                  project -> project.dueDate()
+                                                    .isLessThan(dueDateValue))
+                          .build();
+
+        FilterAdapter adapter = FilterAdapter.of(new DsColumnMapping());
+        Subject<StgProjectId, StgProject> subject = query.subject();
+        Collection<StructuredQuery.Filter> filters =
+                fromPredicate(subject.predicate(), adapter);
+
+        IterableSubject asserted = assertThat(filters);
```
dmdashenkov:
```suggestion
        IterableSubject assertFilters = assertThat(filters);
```

datastore/src/main/java/io/spine/server/storage/datastore/DsIdentifier.java
```diff
@@ -32,19 +32,26 @@
...
+     *
+     * Overrides to declare the method as {@code final}.
+     */
+    @Override
+    public final String value() {
+        return super.value();
     }
 
     @Override
+    @SuppressWarnings("DuplicateStringLiteralInspection")       /* Common term used. */
```
dmdashenkov: Please fix the amount of whitespace.

dmdashenkov (approved):
LGTM with minor comments.

datastore/src/main/java/io/spine/server/storage/datastore/query/FilterAdapter.java
```diff
@@ -24,55 +24,50 @@
...
+    public static FilterAdapter of(ColumnMapping<Value<?>> columnMapping) {
+        checkNotNull(columnMapping);
         return new FilterAdapter(columnMapping);
     }
 
     private FilterAdapter(ColumnMapping<Value<?>> columnMapping) {
         this.columnMapping = columnMapping;
     }
 
+    ColumnMapping<Value<?>> columnMapping() {
```
alexander-yevsyukov, line 57: Please document this package-access method.

alexander-yevsyukov (approved):
LGTM with one comment.

PR #185, opened 2023-08-25
Use Docker-based Datastore emulator via Testcontainers

armiol:
Previously, most of the tests were targeting a local Datastore emulator, started via Gradle's `ExecFork` plugin. It was a decent solution, but it had its downsides:

1. `ExecFork` plugin was not too configurable in terms of wiring it with our execution flow — Gradle complained on its implicit dependencies.

2. Once tests failed for some reason, or, in some cases, if all of them pass, the emulator was not stopped. Such a behaviour is still observed even with the most recent version of `ExecFork` plugin.

3. Apparently, partly due to p.2 we were maintaining an [`ExecFork` fork (so to speak)](https://github.com/SpineEventEngine/gradle-execfork-plugin), which took time to manage. As of now, our fork is really outdated, comparing to `ExecFork`'s `master` branch.

This changeset migrates the repository tests to using a Docker-based Datastore emulator via [`testcontainers/GCloud` module](https://java.testcontainers.org/modules/gcloud/).

Now, to launch a build, one should have a local Docker up and running.

As to GH Actions, its environment provides some container-related functionality, which is apparently sufficient for Testcontainers to work.

Also, some slow tests on catch-up functionality were disabled, as they were is the past. It helps to reduce the build time.


armiol:
@alexander-yevsyukov PTAL.

testutil-gcloud/src/main/java/io/spine/testing/server/storage/datastore/EmulatorContainer.java
```diff
@@ -0,0 +1,107 @@
...
+import org.testcontainers.utility.DockerImageName;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A container running the Emulator image.
+ *
+ * <p>The container is started in its "full consistency" mode.
+ */
+class EmulatorContainer extends DatastoreEmulatorContainer {
```
alexander-yevsyukov: Can we have this class `final`, if we do not want to extend it in the same package?

alexander-yevsyukov (approved):
LGTM, with one minor request.

testutil-gcloud/src/main/java/io/spine/testing/server/storage/datastore/EmulatorContainer.java
armiol: Addressed.

